PR #9, opened 2019-09-19
Fix LGTM alerts

diabonas:
Fix all the current [LGTM alerts](https://lgtm.com/projects/g/tpm2-software/tpm2-pytss/alerts/), which are mainly import-related and very low-hanging fruit.


pdxjohnny:
Awesome! Thank you!

PR #27, opened 2020-03-17
tests: esys: LoadExternal

pdxjohnny:


Superhepper:
It is weird that the 'secrets' module is causing trouble. Because it was introduced in Python 3.6. Right now it can be removed because I used it previously in order to generate a seed in the private section when I did some other LoadExternal tests.

PR #32, opened 2020-05-08
MANIFEST: distribute examples

diabonas:
[`examples/fapi_get_random.py`](https://github.com/tpm2-software/tpm2-pytss/blob/master/examples/fapi_get_random.py) is required by [`tests/test_examples.py`](https://github.com/tpm2-software/tpm2-pytss/blob/master/tests/test_examples.py) and might be helpful for users as well. Examples are only included in the distribution and not installed by Python.


pdxjohnny:
When I added the example I was wondering if you'd say that :) Thanks for the PR!

PR #33, opened 2020-05-15
tests: use ek_cert_less for mssim FAPI tests for tpm2-tss 2.4.1

diabonas:
Since upstream commit https://github.com/tpm2-software/tpm2-tss/commit/e974a73b087c8889a3d138b09fbe71031ed9f8b8 it is necessary to use `ek_cert_less` when testing with the TPM simulator since the EK is now used as a session key by default and the TPM simulator does not provide an EK certificate.

### DIFF
--- a/.ci/download-deps.sh
+++ b/.ci/download-deps.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: BSD-3-Clause
 
-# TPM2 2.4.0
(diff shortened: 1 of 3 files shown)


pdxjohnny:
Awesome! Thanks!

pdxjohnny:
If you ever want to version bump just make a commit like this: https://github.com/tpm2-software/tpm2-pytss/commit/4b04ed563bfa4d1e9593901fe496305d60b29841 then when I merge it'll release the new version on PyPi :)

PR #45, opened 2020-11-13
Fix size_t pointer args

joholl:
Fix two issues.

A) context.py handles function input parameters.
`UINT<n>_T` are special because for their pointer type `UINT<n>_PTR`, the __T_
suffix is omitted. That's not the case with all other types, including
`SIZE_T`, the pointer type of which is _SIZE_T_PTR_.
    
context.py accounts for the `uint<n>_t` special case. However, this code
path is also matches for _size_t_ types, treating them as a special case,
too, when they are in fact not.

B) In some cases, when NULL/None is returned, a cast to a byte array fails. Add a check.


pdxjohnny:
Thank you! I ran into this the other day myself

PR #51, opened 2021-01-11
Fix tpml digest binding

dgptha:
Fix the TPML_DIGEST type binding on the ByteArrayHelper and added test of the binding.

Also fixed the location of the sysconfig dir provided to the package to match the one in the docker image. Otherwise the procedure described in HACKING.md wasn't working.


williamcroberts:
We have completely abandoned swig and have moved to CFFI. The bindings should be fairly complete for esys/ESAPI, however not completely tested.

PR #58, opened 2021-03-05
Add TCTI Ldr

williamcroberts:
- Add the TCTI Ldr library so testing framework can setup either swtpm or tpm_server without blowing up
- Update the tests to use tcti ldr and some more smarts in starting the simulator
- Restructure the code base for simplicity. We don't need N submodules for each portion like the C library, lets just make it one 
monolithic beast.

@pdxjohnny Is the monolithic beast option a great option? My thought is, that if you have python, building all the tss2 bits into one Shared Object shouldn't be a problem. Its not like you would say I don't want FAPI?

@AndreasFuchsSIT and @joholl do you guys have thoughts on this?

Part of the restructure was to make it easy for @joholl to drop his fapi stuff in. He just needs a FAPI.py, the same for SAPI.py, and then we can add in the individual TCTIs, which might be harder...

The individual TCTI's are platform specific. But we could just keep it all at the TCTI Ldr level for now.

Thoughts?


williamcroberts:
One of the other reasons I went this way, is that it greatly simplifies the handling of the header files. We would need to include all the header files, processed so they get along with CFFI for each module, which is possible, but just seemed like overkill.

scripts/prepare_headers.py
```diff
@@ -19,9 +19,13 @@ def prepare(indir, outfile):
         os.path.join(indir, "tss2_tpm2_types.h"), mode="r", encoding="utf-8"
     ).read()
 
+    s += io.open(os.path.join(indir, "tss2_tcti.h"), mode="r", encoding="utf-8").read()
```
pdxjohnny, line 22: This is just another way to do the same thing, I find pathlib to be a "less lines" approach and have been enjoying it, so wanted to share.

```python
    s += pathlib.Path(indir, "tss2_tcti.h").read_text()
```

pdxjohnny:
Reusing the old testing code might help here. It retries to avoid connection issues such as the ones which I believe are the current cause of CI failure. Do you want me to port them?

- https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/simulator.py
- https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/testing.py
- https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/retry.py

williamcroberts:
> Reusing the old testing code might help here. It retries to avoid connection issues such as the ones which I believe are the current cause of CI failure. Do you want me to port them to match this PR?
> 
> * https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/simulator.py
> * https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/testing.py
> * https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/retry.py

If I deconstruct the exception thrown by _chkrc():
```
  File "/workspace/tpm2-pytss/tpm2_pytss/utils.py", line 10, in _chkrc
    raise Exception(rc >> 16, rc & 0xFFFF)
Exception: (10, 21)
```
I get 10 << 16 + 21 = 655381
```bash
tpm2_rc_decode 655381
tcti:Functionality not supported
```

Which makes me think something is wrong on the Docker image.

williamcroberts:
If I was a betting man, I bet we're not seeing the error becuase:
```C
    *handle = dlopen(file_xfrm, RTLD_NOW);
    if (*handle == NULL) {
        LOG_DEBUG("Failed to load TCTI for name \"%s\": %s", file, dlerror());
        return TSS2_TCTI_RC_NOT_SUPPORTED;
    }
```

williamcroberts:
Another reason I am not too worried about per-tcti integration, is that via Tctildr, one can use the no-dl option and then configure a specific set of tctis for very specific environemnts

williamcroberts:
> > Reusing the old testing code might help here. It retries to avoid connection issues such as the ones which I believe are the current cause of CI failure. Do you want me to port them to match this PR?
> > 
> > * https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/simulator.py
> > * https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/testing.py
> > * https://github.com/tpm2-software/tpm2-pytss/blob/2f9ebdf5c3db233107cb0fa69137e70d38b8887e/tpm2_pytss/util/retry.py
> 
> If I deconstruct the exception thrown by _chkrc():
> 
> ```
>   File "/workspace/tpm2-pytss/tpm2_pytss/utils.py", line 10, in _chkrc
>     raise Exception(rc >> 16, rc & 0xFFFF)
> Exception: (10, 21)
> ```
> 
> I get 10 << 16 + 21 = 655381
> 
> ```shell
> tpm2_rc_decode 655381
> tcti:Functionality not supported
> ```
> 
> Which makes me think something is wrong on the Docker image.

Figured it out, the image has swtpm installed, but the Dockerfile installs tss2 2.4, which doesn't have swtpm tcti. So I could add a check into the test code to check for the tcti as well before firing it up.

scripts/prepare_headers.py
williamcroberts: I ended up re-working the regex prepare_headers.py file to more organized and made this change when doing it. As well as added FAPI and cleaned up the testing code a bit.

joholl:
LGTM. Tested with both `mssim` and `swtpm`. The Tctildr and Fapi bindings are also working.

PR #59, opened 2021-03-06
Use pkgconfig to discover include paths

joholl:
Use `pkgconfig` to get the include paths for the libraries. This also works for manually installed libraries (`/usr/local`).

Also, the build script `libtss2_build.py` gave me a TypeError, so I had to move the `debug=True`

I was not to sure how to add `pkgconfig` as a dependency. Would that be in `.ci/Dockerfile.tpm2-tss-python` or `.ci/download-deps.sh`?

**NOTE**: This PR sits on top of #58.

### DIFF
--- a/scripts/libtss2_build.py
+++ b/scripts/libtss2_build.py
@@ -3,7 +3,27 @@ from cffi import FFI
 ffibuilder = FFI()
 
 import os
+import pkgconfig
+import re
 import sys
+
+
+def get_include_paths(library_names):
+    if not isinstance(library_names, list):
+        library_names = [library_names]
+
+    # find search paths for libraries and header via pkg-config
+    header_dirs = set()
+
+    # for manually installed packages, env var PKG_CONFIG_PATH might need to be changed
+    for library_name in library_names:
+        cflags = pkgconfig.cflags(library_name)
+        header_dirs.update(re.findall(r"(?<=-I)\S+", cflags))
+
+    return header_dirs
+
+
+libraries = ["tss2-esys", "tss2-tctildr", "tss2-fapi"]
 
 # Set up the search path so we find prepare_header and other modules
 PATH = os.path.dirname(__file__) if len(os.path.dirname(__file__)) > 0 else os.getcwd()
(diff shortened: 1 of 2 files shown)


scripts/libtss2_build.py
```diff
@@ -3,8 +3,29 @@
...
+
+def get_include_paths(library_names):
+    if not isinstance(library_names, list):
+        library_names = [library_names]
+
+    # find search paths for libraries and header via pkg-config
+    header_dirs = set()
+
+    # include locally installed packages on common distros
+    os.environ["PKG_CONFIG_PATH"] = "/usr/lib/local"
```
williamcroberts, line 19: I don't think we want to clobber PKG_CONFIG_PATH as it could be set by the user. I also don't think we want to add any differing behavior from autotools. In those cases you have to export PKG_CONFIG_PATH or change the config.

williamcroberts (changes requested):
I think this patch will fix your issue:
```diff
diff --git a/.ci/Dockerfile.tpm2-tss-python b/.ci/Dockerfile.tpm2-tss-python
index 699b5adbc974..eda07614653b 100644
--- a/.ci/Dockerfile.tpm2-tss-python
+++ b/.ci/Dockerfile.tpm2-tss-python
@@ -7,7 +7,7 @@ RUN apt-get -y update && \
     python3-pip \
     swig && \
   rm -rf /var/lib/apt/lists/* && \
-  python3 -m pip install -U pip setuptools wheel virtualenv
+  python3 -m pip install -U pip setuptools wheel virtualenv pkgconfig
 
 COPY .ci/download-deps.sh /workspace/tpm2-pytss/.ci/download-deps.sh
```

For error:
```
import pkgconfig
ModuleNotFoundError: No module named 'pkgconfig'
```

scripts/libtss2_build.py
joholl: You're right. Removing this.

At some point, we might want to point out in the build instructions that the user can point to a custom TSS installation via the `PKG_CONFIG_PATH`.

joholl:
I had to add the dependency to `setup.cfg`

https://github.com/tpm2-software/tpm2-pytss/blob/9b9cfabca7aad12114cff25ff379020b8e8d47ad/setup.cfg#L22-L28__

williamcroberts (approved):

PR #63, opened 2021-03-10
Fix test suite

diabonas:
Currently [no tests are being run](https://github.com/tpm2-software/tpm2-pytss/runs/2067384711#step:6:325) because unittest is not able to find the test files. Fix this by renaming them to include a `test_` prefix in the filename, and fix various small issues found in the process that would otherwise lead to test failures.


diabonas:
Tests are now [being run as expected](https://github.com/tpm2-software/tpm2-pytss/runs/2080903570#step:6:6222):
```
Ran 4 tests in 4.044s

OK
```

PR #67, opened 2021-03-11
utils: parse TSS2 return code in exception

whooo:
This is a more or less a cut and paste from the previous bindings.
Useful when debugging and could be used to provide better feedback/error messages from applications to users.

### DIFF
--- a/tpm2_pytss/TSS2_Exception.py
+++ b/tpm2_pytss/TSS2_Exception.py
@@ -1,2 +1,25 @@
+from .types import TPM2_RC
+
+
 class TSS2_Exception(RuntimeError):
-    pass
+    def __init__(self, rc):
+        super(TSS2_Exception, self).__init__(f"TSS2 Library call failed with: 0x{rc:X}")
+        self.rc = rc
+        self.handle = 0
+        self.parameter = 0
+        self.session = 0
+        self.error = 0
+        if self.rc & TPM2_RC.FMT1:
+            self._parse_fmt1()
+        else:
+            self.error = self.rc
+
+    def _parse_fmt1(self):
+        self.error = TPM2_RC.FMT1 + (self.rc & 0x3F)
+
+        if self.rc & TPM2_RC.P:
+            self.parameter = (self.rc & TPM2_RC.N_MASK) >> 8
+        elif self.rc & TPM2_RC.S:
+            self.session = ((self.rc - TPM2_RC.S) & TPM2_RC.N_MASK) >> 8
+        else:
+            self.handle = (self.rc & TPM2_RC.N_MASK) >> 8
--- a/test/test_exception.py
+++ b/test/test_exception.py
@@ -1,0 +1,40 @@
+import unittest
+
+from tpm2_pytss import TSS2_Exception, TPM2_RC
+
+
+class ExceptionTest(unittest.TestCase):
+    def test_non_fmt1(self):
+        exc = TSS2_Exception(TPM2_RC.SESSION_HANDLES)
+        self.assertEqual(exc.rc, TPM2_RC.SESSION_HANDLES)
+        self.assertEqual(exc.error, TPM2_RC.SESSION_HANDLES)
+        self.assertEqual(exc.handle, 0)
+        self.assertEqual(exc.parameter, 0)
+        self.assertEqual(exc.session, 0)
+
+    def test_handle(self):
+        rc = TPM2_RC.TYPE + TPM2_RC.H + TPM2_RC.RC1
+        exc = TSS2_Exception(rc)
+        self.assertEqual(exc.rc, rc)
+        self.assertEqual(exc.error, TPM2_RC.TYPE)
+        self.assertEqual(exc.handle, 1)
+        self.assertEqual(exc.parameter, 0)
+        self.assertEqual(exc.session, 0)
+
+    def test_parameter(self):
+        rc = TPM2_RC.ATTRIBUTES + TPM2_RC.P + TPM2_RC.RC1
+        exc = TSS2_Exception(rc)
+        self.assertEqual(exc.rc, rc)
+        self.assertEqual(exc.error, TPM2_RC.ATTRIBUTES)
+        self.assertEqual(exc.handle, 0)
+        self.assertEqual(exc.parameter, 1)
+        self.assertEqual(exc.session, 0)
+
+    def test_session(self):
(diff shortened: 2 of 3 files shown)


tpm2_pytss/TSS2_Exception.py
```diff
@@ -1,2 +1,24 @@
+from .types import TPM2_RC
+
 class TSS2_Exception(RuntimeError):
-    pass
+    def __init__(self, rc):
+        super(RuntimeError, self).__init__(f"TSS2 Library call failed with: 0x{rc:X}")
```
williamcroberts, line 6: Wy don't we just bind it to rcdecode and get that string back and pass that up?

williamcroberts (commented):
What's the output look like when an exception is thrown?

Having that in the commit message is useful for history.


```
Was like this

Changed it to like this...
```

tpm2_pytss/TSS2_Exception.py
whooo: Right now it's because I haven't gotten my head around cffi so I'm not ready to create bindings for libtss2-rc just yet, I can either mark this as WIP and fix it up when there is bindings for the RC decoder or if it's commited I'll fix fix it when there is bindings.

williamcroberts: I can do it, takes like 2 seconds in an ideal world :-p

whooo:
> What's the output look like when an exception is thrown?
> 
> Having that in the commit message is useful for history.
> 
> ```
> Was like this
> 
> Changed it to like this...
> ```

The output is the same, example:
```
>>> from tpm2_pytss import *
>>> ectx = ESAPI()
>>> ectx.GetRandom(0, session1=0)
ERROR:esys:src/tss2-esys/esys_iutil.c:130:init_session_tab() Error: ESYS_TR is not a session resource. 
ERROR:esys:src/tss2-esys/api/Esys_GetRandom.c:161:Esys_GetRandom_Async() Initialize session resources ErrorCode (0x00070018) 
ERROR:esys:src/tss2-esys/api/Esys_GetRandom.c:72:Esys_GetRandom() Error in async function ErrorCode (0x00070018) 
  File "/home/usertpm2-pytss/tpm2_pytss/ESAPI.py", line 724, in GetRandom
    _chkrc(
  File "/home/usertpm2-pytss/tpm2_pytss/utils.py", line 11, in _chkrc
    raise TSS2_Exception(rc)
tpm2_pytss.TSS2_Exception.TSS2_Exception: TSS2 Library call failed with: 0x70018

```

williamcroberts:
> > What's the output look like when an exception is thrown?
> > Having that in the commit message is useful for history.
> > ```
> > Was like this
> > 
> > Changed it to like this...
> > ```
> 
> The output is the same, example:
> 
> ```
> >>> from tpm2_pytss import *
> >>> ectx = ESAPI()
> >>> ectx.GetRandom(0, session1=0)
> ERROR:esys:src/tss2-esys/esys_iutil.c:130:init_session_tab() Error: ESYS_TR is not a session resource. 
> ERROR:esys:src/tss2-esys/api/Esys_GetRandom.c:161:Esys_GetRandom_Async() Initialize session resources ErrorCode (0x00070018) 
> ERROR:esys:src/tss2-esys/api/Esys_GetRandom.c:72:Esys_GetRandom() Error in async function ErrorCode (0x00070018) 
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/home/usertpm2-pytss/tpm2_pytss/ESAPI.py", line 724, in GetRandom
>     _chkrc(
>   File "/home/usertpm2-pytss/tpm2_pytss/utils.py", line 11, in _chkrc
>     raise TSS2_Exception(rc)
> tpm2_pytss.TSS2_Exception.TSS2_Exception: TSS2 Library call failed with: 0x70018
> ```

Oh it just lets those that catch the exception get at the fields if you need them, which would still be useful. I would add that to the commit message, ie it's intended use, and I would add a test.

williamcroberts:
@joholl FYI I have this PR plumbing it to RC Decode:
  - https://github.com/tpm2-software/tpm2-pytss/pull/68

I thought about creating a routine in rc decode that would return a structure with all the bits decomposed. This way folks don't need to hand jamb through all the masks and stuff. I keep meaning to add this, it would be nice here to just get at em and not re-invent the wheel.

whooo:
@williamcroberts, added some context to the commit message and created tests

tpm2_pytss/TSS2_Exception.py
williamcroberts: @whoo, looks good, you need to change
```python
super(RuntimeError, self).__init__(f"TSS2 Library call failed with: 0x{rc:X}")
```

with:
```python
super(TSS2_Exception, self).__init__(f"TSS2 Library call failed with: 0x{rc:X}")
```
to make LGTM happy.

williamcroberts (changes requested):

whooo:
> ~@joholl~ whoops meant to put @whooo: FYI I have this PR plumbing it to RC Decode:
> 
>     * #68
> 
I saw that, do you want me to rebase this PR over that PR?
> 
> I thought about creating a routine in rc decode that would return a structure with all the bits decomposed. This way folks don't need to hand jamb through all the masks and stuff. I keep meaning to add this, it would be nice here to just get at em and not re-invent the wheel.

Are you talking about adding it to libtss2-rc? Something like Tss2_RC_Unmangle(TSS2_RC, struct *rc_unmasked); ?

williamcroberts:
> > ~@joholl~ whoops meant to put @whooo: FYI I have this PR plumbing it to RC Decode:
> > ```
> > * #68
> > ```
> 
> I saw that, do you want me to rebase this PR over that PR?

No Ill deal with that, since you were first.

> 
> > I thought about creating a routine in rc decode that would return a structure with all the bits decomposed. This way folks don't need to hand jamb through all the masks and stuff. I keep meaning to add this, it would be nice here to just get at em and not re-invent the wheel.
> 
> Are you talking about adding it to libtss2-rc? Something like Tss2_RC_Unmangle(TSS2_RC, struct *rc_unmasked); ?

Yep exactly.

then you can do things like:
```C
if (rc_unmasked.format == 1) {
   rc_unmasked.format_1.layer
   rc_unmasked.format_1.error
   ...
}
```

williamcroberts:
Clean LGTM: https://lgtm.com/projects/g/tpm2-software/tpm2-pytss/rev/pr-b903161de7c0d22387d177f1887cfe98277ed5a7

PR #69, opened 2021-03-12
test: mark FAPI tests as @pytest.mark.forked

diabonas:
Each FAPI test needs to be run in its own subprocess because it [exports the `TSS2_FAPICONF`](https://github.com/tpm2-software/tpm2-pytss/blob/e32432f691eb26aaa31603c0d29427a36fdcbc8f/tpm2_pytss/FAPI.py#L78) (a.k.a. [`FAPI_CONFIG_ENV`](https://github.com/tpm2-software/tpm2-pytss/blob/e32432f691eb26aaa31603c0d29427a36fdcbc8f/tpm2_pytss/FAPI.py#L15)) environment variable to point to its own temporary FAPI configuration. Not isolating the tests in different subprocesses leads to errors like
```
        if self.config is None:
            # Load the currently active fapi-config.json
            config_path = os.environ.get(FAPI_CONFIG_ENV, FAPI_CONFIG_PATH)
>           with open(config_path) as file:
E           FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp3e2_8a4c'
```
because the next test tries to set up the configuration from the environment variable, which points to the old and possibly already deleted temporary configuration. This can be observed in the [latest master CI build](https://github.com/tpm2-software/tpm2-pytss/runs/2097176581#step:6:439).

Mark the FAPI tests as `@pytest.mark.forked` from the pytest-forked/pytest-xdist plugin (which we [already require as a development dependency](https://github.com/tpm2-software/tpm2-pytss/blob/e32432f691eb26aaa31603c0d29427a36fdcbc8f/setup.cfg#L40)) to force them to be run in individual subprocesses regardless of how the user calls pytest.

### DIFF
--- a/test/test_fapi.py
+++ b/test/test_fapi.py
@@ -3,12 +3,13 @@
 SPDX-License-Identifier: BSD-3
 """
 
-import unittest
+import pytest
 
 from tpm2_pytss import *
 from .TSS2_BaseTest import TSS2_FapiTest
 
 
+@pytest.mark.forked
 class TestFapi(TSS2_FapiTest):
     def testProvision(self):
         r = self.fapi.provision()


williamcroberts:
> Each FAPI test needs to be run in its own subprocess because it [exports the `TSS2_FAPICONF`](https://github.com/tpm2-software/tpm2-pytss/blob/e32432f691eb26aaa31603c0d29427a36fdcbc8f/tpm2_pytss/FAPI.py#L78) (a.k.a. [`FAPI_CONFIG_ENV`](https://github.com/tpm2-software/tpm2-pytss/blob/e32432f691eb26aaa31603c0d29427a36fdcbc8f/tpm2_pytss/FAPI.py#L15)) environment variable to point to its own temporary FAPI configuration. Not isolating the tests in different subprocesses leads to errors like
> 
> ```
>         if self.config is None:
>             # Load the currently active fapi-config.json
>             config_path = os.environ.get(FAPI_CONFIG_ENV, FAPI_CONFIG_PATH)
> >           with open(config_path) as file:
> E           FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp3e2_8a4c'
> ```
> 
> because the next test tries to set up the configuration from the environment variable, which points to the old and possibly already deleted temporary configuration. This can be observed in the [latest master CI build](https://github.com/tpm2-software/tpm2-pytss/runs/2097176581#step:6:439).
> 
> Mark the FAPI tests as `@pytest.mark.forked` from the pytest-forked/pytest-xdist plugin (which we [already require as a development dependency](https://github.com/tpm2-software/tpm2-pytss/blob/e32432f691eb26aaa31603c0d29427a36fdcbc8f/setup.cfg#L40)) to force them to be run in individual subprocesses regardless of how the user calls pytest.

I guess I didn't understand the docs, I thought that's what it was doing already. Thanks.

diabonas:
> I thought that's what it was doing already

It does indeed if you run it e.g. as `python3 -m pytest -n $(nproc)` like you suggested in the commit message of d03e6e6ec76d133f8aac6cae6d516fc41f8f1f78, this makes it work even if you call it without any arguments (as long as the necessary pytest plugin is installed).

williamcroberts:
Then why did master CI go down?
https://github.com/tpm2-software/tpm2-pytss/blob/master/.ci/docker.run#L14

In commit 91454565 I updated that command and then in a subsequent commit, d03e6e6ec changed the test class.

diabonas:
My mistake, apparently `-n` doesn't guarantee that each test is run in its own subprocess: I guess (but am not sure, since I haven't looked into the code) it creates multiple subprocesses to parallelise the tests, but reuses them after the previous test has finished. If my assumption is correct, this would explain why it only fails sometimes, in tthe case that a worker which has previously run a FAPI test is reused for another FAPI test.

The actual command line equivalent of `@pytest.mark.forked` would be `--forked` according to the [pytest-forked documentation](https://github.com/pytest-dev/pytest-forked#usage-examples), which can be used on its own or in combination with `-n` to start multiple subprocesses in parallel.

williamcroberts:
> My mistake, apparently `-n` doesn't guarantee that each test is run in its own subprocess: I guess (but am not sure, since I haven't looked into the code) it creates multiple subprocesses to parallelise the tests, but reuses them after the previous test has finished. If my assumption is correct, this would explain why it only fails sometimes, in tthe case that a worker which has previously run a FAPI test is reused for another FAPI test.
> 
> The actual command line equivalent of `@pytest.mark.forked` would be `--forked` according to the [pytest-forked documentation](https://github.com/pytest-dev/pytest-forked#usage-examples), which can be used on its own or in combination with `-n` to start multiple subprocesses in parallel.

My research has led me to a similar conclusion. But as always thanks for the patches!

williamcroberts:
I think I like the forked built into the code rather than the problem than the purest command so it'll always be right, less prone the user error.

PR #73, opened 2021-03-29
Support better pcr select

williamcroberts:


williamcroberts:
@whooo native python types heading your way. Essentially we hide things that are scalar values and wrap compound things with an actual class.

tpm2_pytss/types.py
```diff
@@ -4,8 +4,10 @@
 
 from ._libtpm2_pytss import ffi, lib
 
+from tpm2_pytss.utils import CLASS_INT_ATTRS_from_string, TPM2B_unpack
```
pdxjohnny, line 7: Mixed import styles, relative on line 5, absolute on like 7

```suggestion
from .utils import CLASS_INT_ATTRS_from_string, TPM2B_unpack
```

tpm2_pytss/types.py
```diff
@@ -732,399 +743,715 @@ class TPMA_MEMORY(int):
...
+    @staticmethod
+    def parse(selection):
 
-class TPMS_EMPTY(object):
-    def __new__(self):
-        return ffi.new("TPMS_EMPTY *")
+        if selection is None or len(selection) == 0:
+            return TPMS_PCR_SELECTION()
 
+        hunks = [x.strip() for x in selection.split(":")]
```
pdxjohnny, line 1329: You might want to check for the presence of ":" before splitting on it, and similarly raise an exception. Users might get an unhelpful error message if they didn't supply a value with a ":".

tpm2_pytss/types.py
```diff
@@ -732,399 +743,715 @@ class TPMA_MEMORY(int):
...
+    @staticmethod
+    def parse(selections):
 
+        if selections is None or len(selections) == 0:
+            return TPML_PCR_SELECTION()
 
-class TPM2B_SYM_KEY(object):
-    def __new__(self):
-        return ffi.new("TPM2B_SYM_KEY *")
+        selectors = selections.split("+")
```
pdxjohnny, line 1122: Similar to other comment, You might want to check for the presence of "+" before splitting on it, and similarly raise an exception to the one below in the bounds checking. Users might get an unhelpful error message if they didn't supply a value with a "+".

tpm2_pytss/types.py
```diff
@@ -732,399 +743,715 @@ class TPMA_MEMORY(int):
...
+            return tipe.cname[len(tipe.kind) + 1 :]
+
+        return tipe.cname
+
+    @staticmethod
+    def _init_cdata(child, cdata):
+
+        init_data = cdata
+        if cdata is None:
+            init_data = ffi.new(f"{child.__class__.__name__} *")
```
pdxjohnny, line 764: > I'm not entirely clear on why this isn't just a part of the base class's `__init__`, or called within it. `self.__class__.__name__` would be the name of the subclass if this was a regular method and called from the base class `__init__`. 
> Seems like that would eliminate the repitiion of
> ```python
>     def __init__(self, _cdata=None):
>         init_data = TPM_OBJECT._init_cdata(self, _cdata)
>         super().__init__(init_data)
> ```

pdxjohnny (commented):
Nice

tpm2_pytss/types.py
williamcroberts: Darn IDE

tpm2_pytss/types.py
williamcroberts: Yep, sounds good, here and throughout on splits from user data

tpm2_pytss/types.py
williamcroberts: My reason was I didn't think this worked as expected:
```python
class FOO(object):
    def __init__(self):
        print(self.__class__.__name__)

class BAR(FOO):
    def __init__(self):
        super().__init__()

f = FOO()
b = BAR()
```
```bash
python3 ~/a.py
FOO
BAR
```
But it does, so I can get rid of all these and just dynamically create them.

tpm2_pytss/types.py
williamcroberts: Theirs a chance only one bank is specified and no + between them, so we want to make sure we handle and test that case. Good catch.

tpm2_pytss/types.py
williamcroberts: Theirs a raise below it, we MUST have 2 chunks. So I can make that exception better.

PR #76, opened 2021-04-02
types.py: support setting struct/union fields during initialization

whooo:
This enables things like TPMS_NV_PUBLIC(nameAlg=TPM2_ALG.SHA256)

### DIFF
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -744,12 +744,23 @@ class TPMA_MEMORY(int):
 
 
 class TPM_OBJECT(object):
-    def __init__(self, _cdata=None):
+    def __init__(self, _cdata=None, **kwargs):
 
         if _cdata is None:
             _cdata = ffi.new(f"{self.__class__.__name__} *")
 
         self._cdata = _cdata
+
+        tipe = ffi.typeof(_cdata)
+        if tipe.kind == "pointer":
+            tipe = tipe.item
+        fields = [x[0] for x in tipe.fields]
+        for k, v in kwargs.items():
+            if k not in fields:
+                raise AttributeError(
+                    f"{self.__class__.__name__} has no field by the name of {k}"
+                )
+            self.__setattr__(k, v)
 
     @staticmethod
     def _fixup_classname(tipe):
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -754,6 +754,10 @@ class TPM_OBJECT(object):
         tipe = ffi.typeof(_cdata)
         if tipe.kind == "pointer":
             tipe = tipe.item
+        if tipe.cname != self.__class__.__name__:
+            raise TypeError(
+                f"Unexpected _cdata type {tipe.cname}, expected {self.__class__.__name__}"
+            )
         fields = [x[0] for x in tipe.fields]
         for k, v in kwargs.items():
             if k not in fields:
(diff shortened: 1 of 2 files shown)


tpm2_pytss/types.py
```diff
@@ -744,13 +744,24 @@ class TPMA_MEMORY(int):
...
 
+        tipe = ffi.typeof(_cdata)
+        if tipe.kind == "pointer":
+            tipe = tipe.item
+        fields = [x[0] for x in tipe.fields]
+        for k, v in kwargs.items():
+            if k not in fields:
+                raise AttributeError(
+                    f"{self.__class__.__name__} has no field by the name of {k}"
+                )
```
williamcroberts, line 762: Couldn't this just be a if key not in self._cdata?
Tests?
What do we map type to in a TPMT_PUBLIC since type is reserved? It looks like Python style guide is to use the name with an underscore, ie `type_`

FYI I have some work here similarly related:
https://github.com/williamcroberts/tpm2-pytss/pull/new/add-int-type-parsers

whooo: key in _cdata doesn't work, at least not for me, and I want make sure that it's an actual struct/union field, not an attribute of the ffi object.
Tests are coming.
type isn't really reserved, so having an attribute called type is OK, it seems to be bad practice, but as long as it's only attributes of TPMS_ATTEST and TPMT_PUBLIC my opinion is that it's OK.
```
>>> pub = TPMT_PUBLIC(type=TPM2_ALG.ECC)
>>> print(pub.type)
35
```

whooo: Test added now

whooo:
While on the subject I'm looking into the following next:
Making the size and count attributes readonly
Creating iterators and __getitem__/__setitem__ for TPML_OBJECT
Creating TPM2B_OBJECT for all TPM2B structs with size + buffer, with iterators and returning bytes when getting the buffer attribute
Creating bindings for libtss2-mu and creating instance(class methods for that

tpm2_pytss/types.py
williamcroberts: > key in _cdata doesn't work, at least not for me, and I want make sure that it's an actual struct/union field, not an attribute of the ffi object.
> Tests are coming.
> type isn't really reserved, so having an attribute called type is OK, it seems to be bad practice, but as long as it's only attributes of TPMS_ATTEST and TPMT_PUBLIC my opinion is that it's OK.
> 
> ```
> >>> pub = TPMT_PUBLIC(type=TPM2_ALG.ECC)
> >>> print(pub.type)
> 35
> ```

type isn't reserved in the sense that it's in pythons list of reserved words, however you will override `type()` locally. Which means if we ever need normal type() we can encounter some weirdness. In this particular case, since it's getting unpacked from kwargs it would be fine, but if things change it could cause weird issues.

williamcroberts:
@whoo FYI I have this PR which is similar to this one, but for the int types:
  - https://github.com/tpm2-software/tpm2-pytss/pull/79

It also has a test for TPM2B_PUBLIC setting parts via constructor.

I am going to add a parse to the TPMT_PUBLIC routine, so you can do something like tpm2-tools style strings to get TPMT_PUBLIC's populated.

williamcroberts:
> While on the subject I'm looking into the following next:
> Making the size and count attributes readonly
> Creating iterators and **getitem**/**setitem** for TPML_OBJECT
> Creating TPM2B_OBJECT for all TPM2B structs with size + buffer, with iterators and returning bytes when getting the buffer attribute
> Creating bindings for libtss2-mu and creating instance(class methods for that

Sounds great.

PR #90, opened 2021-04-10
types.py: add TPM2B_OBJECT to handle size + buffer structures

whooo:
* Makes the size attribute read only
* Keeps track of buffer size when buffer is changed
* Makes buffer data inmutable
* Makes the class indexable, iterable and sliceable

I had to change unrelated tests as they where using TPM2B_DIGEST

### DIFF
--- a/test/test_types.py
+++ b/test/test_types.py
@@ -263,16 +263,13 @@ class TypesTest(unittest.TestCase):
         )
 
     def test_TPM_OBJECT_init(self):
-        digest = TPM2B_DIGEST(size=4, buffer=b"test")
-
-        self.assertEqual(digest.size, 4)
-        b = digest.buffer
-        self.assertEqual(b, b"test")
+        pub = TPM2B_PUBLIC(publicArea=TPMT_PUBLIC(nameAlg=TPM2_ALG.SHA256))
+        self.assertEqual(pub.publicArea.nameAlg, TPM2_ALG.SHA256)
 
         with self.assertRaises(
-            AttributeError, msg="TPM2B_DIGEST has no field by the name of badfield"
+            AttributeError, msg="TPM2B_PUBLIC has no field by the name of badfield"
         ):
-            TPM2B_DIGEST(badfield=1)
+            TPM2B_PUBLIC(badfield=1)
 
     def test_TPM_OBJECT_init_cdata(self):
         with self.assertRaises(
@@ -574,7 +571,6 @@ class TypesTest(unittest.TestCase):
 
     def test_marshal(self):
         pb = TPM2B_PUBLIC()
-        pb.publicArea.authPolicy.size = 8
         pb.publicArea.authPolicy.buffer = b"password"
         b = pb.publicArea.authPolicy.Marshal()
         self.assertEqual(b, b"\x00\x08password")
@@ -1037,20 +1033,37 @@ class TypesTest(unittest.TestCase):
         p = str(x.userAuth)
         self.assertEqual(p, binascii.hexlify("password".encode()).decode())
 
-    def test_TPM2B_NAME(self):
-        name = binascii.unhexlify(
-            "000b34e9133541f7874f5ec2cd867b873b05cea0c1677b0cc7d740988e999bee5450"
-        )
-        print(len(name))
-        x = TPM2B_NAME(name)
-        self.assertEqual(x.size, 34)
-        n = bytes(x)
-        self.assertEqual(len(n), x.size)
-        self.assertEqual(n, name)
+    def test_TPM2B_SIMPLE_OBJECT(self):
+        bob = b"bunchofbytes"
+        dig = TPM2B_NAME(bob)
+        self.assertEqual(dig.name, bob)
+        self.assertEqual(len(dig), len(bob))
 
-    def test_TPM2B_PUBLIC_fail_from_bytes(self):
+        for i in range(0, len(dig)):
+            self.assertEqual(dig[i], bob[i])
+
+        with self.assertRaises(IndexError):
+            dig[len(dig)]
+
+        self.assertEqual(dig[0:3], b"bun")
+
+        self.assertEqual(dig[60:64], b"")
+
         with self.assertRaises(TypeError):
-            TPM2B_PUBLIC(b"0011223344556677889900")
+            dig["str"]
+
+        i = 0
+        for b in dig:
+            self.assertEqual(b, bob[i])
+            i = i + 1
+
+        b = bytes(dig)
+        self.assertEqual(b, bob)
+
+        with self.assertRaises(AttributeError):
+            dig.size = 1
+        with self.assertRaises(TypeError):
+            dig.name[0] = b"\x00"
 
 
 if __name__ == "__main__":
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -6,7 +6,6 @@ from ._libtpm2_pytss import ffi, lib
 
 from tpm2_pytss.utils import (
     CLASS_INT_ATTRS_from_string,
-    TPM2B_unpack,
     _chkrc,
     fixup_cdata_kwargs,
     cpointer_to_ctype,
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -920,65 +920,52 @@ class TPM_OBJECT(object):
         return cls(_cdata=_cdata), offset[0]
 
 
-class TPM2B_OBJECT(TPM_OBJECT):
-    def __init__(self, _cdata=None, **kwargs):
-        _cdata, kwargs = fixup_cdata_kwargs(self, _cdata, kwargs)
-        super().__init__(_cdata, **kwargs)
+class TPM2B_SIMPLE_OBJECT(TPM_OBJECT):
+    @classmethod
+    def _get_bytefield(cls):
+        tipe = ffi.typeof(f"{cls.__name__}")
+        for f in tipe.fields:
+            if f[0] != "size":
+                return f[0]
 
-        # if their is nothing to update OR the caller passed both the
-        # buffer AND the size, their is nothing to do.
-        if len(kwargs) == 0 or len(kwargs) == 2:
-            return
-
-        # we don't set size on nested compound TPM2B structures, like TPM2B_PUBLIC
-        tipe = cpointer_to_ctype(_cdata)
-        field_name = next((v[0] for v in tipe.fields if v[0] != "size"), None)
-
-        field_cdata = self._cdata.__getattribute__(field_name)
-        tipe = ffi.typeof(field_cdata)
-        if tipe.kind != "array":
-            return
-
-        key = [*kwargs][0]
-        value = kwargs[key]
-        if len(kwargs) == 0:
-            return
-
-        # update the len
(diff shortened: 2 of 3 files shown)


whooo:
I'll rebase on master if everything looks good to you

whooo:
ok, I now see there is already an commit for this...

whooo:
@williamcroberts, so this to a large degree overwrites your work, sorry about that, I was almost done with this version.
I skipped all TPM2B structures which aren't a size + a byte array, as size seems ignored by the API for those (I never set the correct size for publicArea in TPM2B_PUBLIC for example).

I wanted to make the byte array immutable, but memoryview.toreadonly() requires python 3.8 or later.
It also supports the structures that have another field then buffer (TPM2B_NAME for example)

There is also a small change to the API, _cdata doesn't come first now, so it needs to be specified (see the change in ESAPY.py), as I think it's useful to be able to to TPM2B_DIGEST(bunch_of_bytes)

williamcroberts:
> @williamcroberts, so this to a large degree overwrites your work, sorry about that, I was almost done with this version.
> I skipped all TPM2B structures which aren't a size + a byte array, as size seems ignored by the API for those (I never set the correct size for publicArea in TPM2B_PUBLIC for example).

Yeah the general paradigm is that callers intentionally leave that to zero, and then libmu figures out the right size.

> 
> I wanted to make the byte array immutable, but memoryview.toreadonly() requires python 3.8 or later.
> It also supports the structures that have another field then buffer (TPM2B_NAME for example)

I thought the current tip of master does that, I had it look for count/size and then the other field is the one you care about.

> 
> There is also a small change to the API, _cdata doesn't come first now, so it needs to be specified (see the change in ESAPY.py), as I think it's useful to be able to to TPM2B_DIGEST(bunch_of_bytes)

I just figured it out by type of data

tpm2_pytss/types.py
```diff
@@ -1460,15 +1463,15 @@ class TPM2B_ATTEST(TPM2B_OBJECT):
     pass
 
 
-class TPM2B_CONTEXT_DATA(TPM2B_OBJECT):
+class TPM2B_CONTEXT_DATA(TPM_OBJECT):
```
williamcroberts, line 1466: Why are we no longer inheriting? These are TPM2B structures.

whooo: Seems I looked at the wrong structure, fixed it now.

whooo:
> > @williamcroberts, so this to a large degree overwrites your work, sorry about that, I was almost done with this version.
> > I skipped all TPM2B structures which aren't a size + a byte array, as size seems ignored by the API for those (I never set the correct size for publicArea in TPM2B_PUBLIC for example).
> 
> Yeah the general paradigm is that callers intentionally leave that to zero, and then libmu figures out the right size.
> 
> > I wanted to make the byte array immutable, but memoryview.toreadonly() requires python 3.8 or later.
> > It also supports the structures that have another field then buffer (TPM2B_NAME for example)
> 
> I thought the current tip of master does that, I had it look for count/size and then the other field is the one you care about.
I was thinking about that only "buffer" as key is being checked in `__getattribute__` and checking for TPM2B_DIGEST ini __init__
> 
> > There is also a small change to the API, _cdata doesn't come first now, so it needs to be specified (see the change in ESAPY.py), as I think it's useful to be able to to TPM2B_DIGEST(bunch_of_bytes)
> 
> I just figured it out by type of data

I see that now, will revert to that part

tpm2_pytss/types.py
```diff
@@ -1516,7 +1517,7 @@ class TPM2B_NAME(TPM2B_OBJECT):
     pass
 
 
-class TPM2B_NV_PUBLIC(TPM2B_OBJECT):
+class TPM2B_NV_PUBLIC(TPM_OBJECT):
```
williamcroberts, line 1520: I guess it's simpler to avoid the complex TPM2B's then handling it in the TPM2B_OBJECT class itself. Can we rename TPM2B_OBJECT to TPM2B_SIMPLE_OBJECT to better clarify that only simple, non compound TPM2B's should inherit from it.

test/test_types.py
```diff
@@ -584,8 +580,7 @@ def test_unmarshal(self):
         d, offset = TPM2B_DIGEST.Unmarshal(buf)
         self.assertEqual(offset, 7)
         self.assertEqual(d.size, 5)
-        db = d.buffer
-        self.assertEqual(db, b"test1")
+        self.assertEqual(d.buffer, b"test1")
```
williamcroberts, line 583: Make this cleanup a separate commit.

williamcroberts (changes requested):
This looks a lot better, just some small changes.

tpm2_pytss/types.py
```diff
@@ -921,54 +921,55 @@ def Unmarshal(cls, buf):
...
 
     def __bytes__(self):
-
-        _cdata = self._cdata
-
-        # not everything uses .buffer, for instance TPM2B_NAME uses .name
-        tipe = cpointer_to_ctype(_cdata)
-        field_name = next((v[0] for v in tipe.fields if v[0] != "size"), None)
-
-        buffer = TPM2B_unpack(_cdata, n=field_name)
```
williamcroberts, line 964: TPM2B_unpack is no longer used remove from imports

tpm2_pytss/types.py
```diff
@@ -921,54 +921,55 @@ def Unmarshal(cls, buf):
...
-
-        # update the len
-        _cdata.size = len(value)
+    def __setattr__(self, key, value):
+        _bytefield = type(self)._get_bytefield()
+        if key == "size":
+            raise AttributeError(f"{key} is read only")
+        super().__setattr__(key, value)
+        if key == _bytefield:
+            self._cdata.size = len(value)
```
williamcroberts, line 941: LGTM now complains in the __init__ for TPM_OBJECT that:
`NEWCall to self.__setattr__ in __init__ method, which is overridden by method TPM2B_OBJECT.__setattr__.`

We should probably have the caller in TPM_OBJECT call:
_cdata.__setattr_(k, v) to avoid that.

williamcroberts (changes requested):
A few things from LGTM. Id like to pull this in as it blocks some ESYS testing, any chance you can make these changes soon, or can I just edit your commit?

test/test_types.py
```diff
@@ -263,16 +263,13 @@ def test_TPM2B_PUBLIC(self):
...
-        self.assertEqual(b, b"test")
+        pub = TPM2B_PUBLIC(publicArea=TPMT_PUBLIC(nameAlg=TPM2_ALG.SHA256))
+        self.assertEqual(pub.publicArea.nameAlg, TPM2_ALG.SHA256)
 
         with self.assertRaises(
-            AttributeError, msg="TPM2B_DIGEST has no field by the name of badfield"
+            AttributeError, msg="TPM2B_PUBLIC has no field by the name of badfield"
         ):
-            TPM2B_DIGEST(badfield=1)
+            TPM2B_PUBLIC(badfield=1)
```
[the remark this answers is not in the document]
whooo: As I changed the init method for TPM2B_DIGEST I wanted to make sure that the this test was for the init method in TPM_OBJECT

tpm2_pytss/types.py
whooo: Sounds good, will add it to the next pusj

tpm2_pytss/types.py
whooo: Using `_cdata.__setattr__(k, v)` in TPM_OBJECTs `__init__` breaks some functionality, then you need to pass cdata for everything

williamcroberts: > Using `_cdata.__setattr__(k, v)` in TPM_OBJECTs `__init__` breaks some functionality, then you need to pass cdata for everything

Ahh Ok, according to https://lgtm.com/rules/10030085/, we want the base class fully initialized before calling to things that are overridden by the subclass. Perhaps we want TPM_OBJECT.__setattr__(self, k, v), so we specifically call the base class setattr here?

whooo: I wonder something like `super().__init__(_cdata)` and then loop over kwargs and running `super().__setattr__(k, v)` in TPM2B_SIMPLE_OBJECT might work instead

whooo: test_TPM2B_AUTH_bad_fields fails then

tpm2_pytss/types.py
whooo: I need to touch ESAPI.py as well when removing TPM2B_unpack, so will make another commit (under this PR) for that

tpm2_pytss/types.py
whooo: Found an ugly hack, lets see if LGTM will accept it

williamcroberts: > Found an ugly hack, lets see if LGTM will accept it

Yeah I thought multiple init calls would trigger a warning, yep:
```
1 for Multiple calls to `__init__` during object initialization
1 for Unused import
1 for `__init__` method calls overridden method
```

On the plus side we got a different result :-p

williamcroberts: > test_TPM2B_AUTH_bad_fields fails then

Is that the only test that fails? Maybe some logic change to the base class? It must be calling setattr on the object over cdata which will work. Perhaps in the constructor we limit the values that can be setattr'd to valid fields in the cdata?

whooo: yeah, exactly, the field check is in the init method, so perhaps a `_setfield` method in TPM_OBJECT that only handles cdata fields and use that in its init method and  `__setattr__`

whooo: yeah, the fields check is in `__init__`,

tpm2_pytss/types.py
```diff
@@ -920,55 +920,52 @@ def Unmarshal(cls, buf):
...
-
-        # if their is nothing to update OR the caller passed both the
-        # buffer AND the size, their is nothing to do.
-        if len(kwargs) == 0 or len(kwargs) == 2:
-            return
-
-        key = [*kwargs][0]
-        value = kwargs[key]
-        if len(kwargs) == 0:
-            return
```
williamcroberts, line 941: I'm a big fan of programming via the delete key.

williamcroberts (changes requested):
LGTM I can handle the conflicts, or do you? We can fixup any other nits later.

williamcroberts (approved):

whooo:
> LGTM I can handle the conflicts, or do you? We can fixup any other nits later.

I''ll take care of that, need to fix the TPM2B_unpack import anyway, but that will happen tomorrow (CET)

williamcroberts:
> > LGTM I can handle the conflicts, or do you? We can fixup any other nits later.
> 
> I''ll take care of that, need to fix the TPM2B_unpack import anyway, but that will happen tomorrow (CET)

Okay, ill freeze master and wait for that, this way ill rebase on you.

whooo:
LGTM-bot is still not happy about `__init__` :-/

Other then that everything is fixed, perhaps the warning is acceptable?

williamcroberts:
> LGTM-bot is still not happy about `__init__` :-/
> 
> Other then that everything is fixed, perhaps the warning is acceptable?

Yeah, we'll get it later.

PR #92, opened 2021-04-12
Support setting byte buffers in TPM2B if supported

williamcroberts:
For TPM2B types where the sub field is only a buffer, allow constructors to instantiate it from bytes. This will be for things like TPM2B_DIGEST and TPM2B_NAME structures. This will also fail on compound structures like TPM2B_PUBLIC.

@ Whoo what about this over parts of 90? You can make size RO ontop of this PR?

### DIFF
--- a/test/test_types.py
+++ b/test/test_types.py
@@ -1048,6 +1048,10 @@ class TypesTest(unittest.TestCase):
         self.assertEqual(len(n), x.size)
         self.assertEqual(n, name)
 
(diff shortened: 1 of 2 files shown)


williamcroberts:
I think #90 over this, lets see how that one progresses. It avoids the need to filter out complex TPM2B's by simply not inheriting from TPM2B_OBJECT if its a complex TPM2B like TPM2B_PUBLIC.

PR #96, opened 2021-04-15
Drop codecov token

williamcroberts:
- Drop codecov token from the secrets, its not needed.
- Cleanup exports into the env variable for Github Actions.
- Make the run.sh script use an argument over magic env variable.


williamcroberts:
The funny thing is, we weren't even uploading coverage reports. We have CODECOV_TOKEN exposed for naught.
The CI has em:
 - https://github.com/tpm2-software/tpm2-pytss/pull/96/checks?check_run_id=2354853308#step:6:596

CodeCov:
 - No coverage uploaded
 - https://app.codecov.io/gh/tpm2-software/tpm2-pytss/pulls

PR #100, opened 2021-04-16
ci: move codecov upload into container

williamcroberts:
Codecov is uploading, but the results file on the server is not valid.
Suggested fixes hinted that the path names were incorect. I attempted
various fixes, including useg the suggested fixes stanza in the
codecov.yml file. However, thay feature, while documented, seems
unimplemented in both the bash and python uploaders[1]. I also tried
changing the mount point internal to the container to /tpm2-pytss, but
that didn't work either. I even tried a script to run before upload to
manually modify the .coverage SQlite3 DB produced via coverage enabled
python runs, but to no avail.

Just run all the tests in the docker container and leave the dist upload
external.

[1]: https://github.com/codecov/codecov-python/issues/318


williamcroberts:
@pdxjohnny YEAH coverage report of 80%

PR #102, opened 2021-04-16
more ci changes

williamcroberts:
- create a publish package job and limit it to PUSH TAG events.
- set the SSH account for publishing docs to github pages branch to not @pdxjohnny.
- make testing the docs for PR's and PUSH requests only run the sphinx command (scripts/docs.sh)
- make publishing docs it's own job and limit to PUSH events.


williamcroberts:
@pdxjohnny here is my dilema, how can I test that the publish jobs for both:
1. docs
2. tagged releases

work before uploading?

Issue #88, opened 2021-04-09
TPM2B_PUBLIC and TPM2B_SENSITIVE from PEM files


williamcroberts:
create TPM2B_PUBLIC.fromPEM() and TPM2B_SENSITIVE.fromPEM routines to convert those data types for things like LoadExternal. Also support a fromAES? How do we want to handle symmetric keys? Should we just be able to pass a pem to the existing parse routines?

williamcroberts:
@whooo what do you think about how you'd like to see this interface? Any chance you got bandwith to work on it while I continue testing other parts of ESAPI bindings?

whooo:
I could work on this (I already have code that does), but it creates some questions, how heavy can we be on dependencies? I like python-cryptography ( https://cryptography.io/en/latest/ ), but least an ASN.1 parser module is required, also should there be any default object attributes for the public part?

I'll look into adding a toPEM method for TPM2B_PUBLIC as well.

williamcroberts:
> I could work on this (I already have code that does), but it creates some questions, how heavy can we be on dependencies? I like python-cryptography ( https://cryptography.io/en/latest/ ), but least an ASN.1 parser module is required, also should there be any default object attributes for the public part?
> 
> I'll look into adding a toPEM method for TPM2B_PUBLIC as well.

Use whatever makes sense, I would use Python Cryptography as that seems to be the defacto standard. I wouldn't worry about dependencies, we are already in Python. Just make sure we update the setup.cfg to include them so pip install doesn't miss anything.

For object attributes, I would require the caller to specify them or manually set them in the returned TPM2B_XXX object.

For toPEM routines, we want to use the format for private objects that tpm2-openssl uses:
  - https://github.com/tpm2-software/tpm2-openssl/blob/master/src/tpm2-provider-pkey.c#L12

That structure is what needs to be duplicated, it gets written as an ASN1.

As always, thanks. I appreciate your help. I'll have to set up an Authors file and make sure you end up in it.

whooo:
I might have to create two to methods, I see a use of just getting the public key as a standard PEM file, so you can for example verify a signature from the TPM using the openssl CLI, so perhaps toPEM for a standard rsa public key and toTSSKey for the tpm2-opemssl format?

I might add support for loading the public key from a certificate as well, using fromPEM for TPM2B_PUBLIC.

whooo:
[...]

PR #106, opened 2021-04-17
tsskey.py: add class to handle tpm2-openssl keys

whooo:
This is currently only basic mock code to show how I think TSS-keys could be handled.

Related to https://github.com/tpm2-software/tpm2-pytss/issues/88

### DIFF
--- a/tpm2_pytss/tsskey.py
+++ b/tpm2_pytss/tsskey.py
@@ -1,0 +1,253 @@
+"""
+SPDX-License-Identifier: BSD-3
+"""
+
+import warnings
+from ._libtpm2_pytss import lib
+from .types import *
+from asn1crypto.core import ObjectIdentifier, Sequence, Boolean, OctetString, Integer
+from asn1crypto import pem
+
+
+parent_rsa_template = TPMT_PUBLIC(
+    type=TPM2_ALG.RSA,
+    nameAlg=TPM2_ALG.SHA256,
+    objectAttributes=TPMA_OBJECT.USERWITHAUTH
+    | TPMA_OBJECT.RESTRICTED
+    | TPMA_OBJECT.DECRYPT
+    | TPMA_OBJECT.NODA
+    | TPMA_OBJECT.FIXEDTPM
+    | TPMA_OBJECT.FIXEDPARENT
+    | TPMA_OBJECT.SENSITIVEDATAORIGIN,
+    authPolicy=b"",
+    parameters=TPMU_PUBLIC_PARMS(
+        rsaDetail=TPMS_RSA_PARMS(
+            symmetric=TPMT_SYM_DEF_OBJECT(
+                algorithm=TPM2_ALG.AES,
+                keyBits=TPMU_SYM_KEY_BITS(aes=128),
+                mode=TPMU_SYM_MODE(aes=TPM2_ALG.CFB),
+            ),
+            scheme=TPMT_RSA_SCHEME(scheme=TPM2_ALG.NULL),
+            keyBits=2048,
+            exponent=0,
+        ),
+    ),
+)
+
+parent_ecc_template = TPMT_PUBLIC(
+    type=TPM2_ALG.ECC,
+    nameAlg=TPM2_ALG.SHA256,
+    objectAttributes=TPMA_OBJECT.USERWITHAUTH
+    | TPMA_OBJECT.RESTRICTED
+    | TPMA_OBJECT.DECRYPT
+    | TPMA_OBJECT.NODA
+    | TPMA_OBJECT.FIXEDTPM
+    | TPMA_OBJECT.FIXEDPARENT
+    | TPMA_OBJECT.SENSITIVEDATAORIGIN,
+    authPolicy=b"",
+    parameters=TPMU_PUBLIC_PARMS(
+        eccDetail=TPMS_ECC_PARMS(
+            symmetric=TPMT_SYM_DEF_OBJECT(
+                algorithm=TPM2_ALG.AES,
+                keyBits=TPMU_SYM_KEY_BITS(aes=128),
+                mode=TPMU_SYM_MODE(aes=TPM2_ALG.CFB),
+            ),
+            scheme=TPMT_ECC_SCHEME(scheme=TPM2_ALG.NULL),
+            curveID=TPM2_ECC.NIST_P256,
+            kdf=TPMT_KDF_SCHEME(scheme=TPM2_ALG.NULL),
+        ),
+    ),
+)
+
+rsa_template = TPMT_PUBLIC(
+    type=TPM2_ALG.RSA,
+    nameAlg=TPM2_ALG.SHA256,
+    objectAttributes=TPMA_OBJECT.USERWITHAUTH
+    | TPMA_OBJECT.SIGN_ENCRYPT
+    | TPMA_OBJECT.DECRYPT
+    | TPMA_OBJECT.NODA
+    | TPMA_OBJECT.FIXEDTPM
+    | TPMA_OBJECT.FIXEDPARENT
+    | TPMA_OBJECT.SENSITIVEDATAORIGIN,
+    authPolicy=b"",
+    parameters=TPMU_PUBLIC_PARMS(
+        rsaDetail=TPMS_RSA_PARMS(
+            symmetric=TPMT_SYM_DEF_OBJECT(algorithm=TPM2_ALG.NULL),
+            scheme=TPMT_RSA_SCHEME(scheme=TPM2_ALG.NULL),
+        ),
+    ),
+)
+
+ecc_template = TPMT_PUBLIC(
+    type=TPM2_ALG.ECC,
+    nameAlg=TPM2_ALG.SHA256,
+    objectAttributes=TPMA_OBJECT.USERWITHAUTH
+    | TPMA_OBJECT.SIGN_ENCRYPT
+    | TPMA_OBJECT.NODA
+    | TPMA_OBJECT.FIXEDTPM
+    | TPMA_OBJECT.FIXEDPARENT
+    | TPMA_OBJECT.SENSITIVEDATAORIGIN,
+    authPolicy=b"",
+    parameters=TPMU_PUBLIC_PARMS(
+        eccDetail=TPMS_ECC_PARMS(
+            symmetric=TPMT_SYM_DEF_OBJECT(algorithm=TPM2_ALG.NULL),
+            scheme=TPMT_ECC_SCHEME(scheme=TPM2_ALG.NULL),
+            kdf=TPMT_KDF_SCHEME(scheme=TPM2_ALG.NULL),
+        ),
+    ),
+)
+
+loadablekey_oid = ObjectIdentifier("2.23.133.10.1.3")
+
+# BooleanOne is used to encode True in the same way as tpm2-tss-engine
+class BooleanOne(Boolean):
+    def set(self, value):
+        self._native = bool(value)
+        self.contents = b"\x00" if not value else b"\x01"
+        self._header = None
+        if self._trailer != b"":
+            self._trailer = b""
+
+
+class TSSPrivKey(object):
+    class _tssprivkey_der(Sequence):
+        _fields = [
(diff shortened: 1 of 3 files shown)


tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,20 @@
+"""
+SPDX-License-Identifier: BSD-3
+"""
+
+
+class TSSKey(object):
+    def __init__(self, private, public, emptyAuth=True, parent=lib.TPM2_RH_OWNER):
```
williamcroberts, line 7: No public, public's not needed and uses the existing TPM2B_PUBLIC and PEM formats. `password=None`, and then detect in the code if password is emptyAuth or not.

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,20 @@
+"""
+SPDX-License-Identifier: BSD-3
+"""
+
+
+class TSSKey(object):
```
williamcroberts, line 6: TSS2PrivateKey or something to make it clear that its only for the private PEM files.

tpm2_pytss/tsskey.py
whooo: But the format contains the public key, so it's needed for serialization and will be read during deserialization.

I rather see password at load time, so it can be set directly on the ESYS_TR handle, I suspect that those who would use the class directly (not using fromPEM) will do so to call toPEM directly afterwards

tpm2_pytss/tsskey.py
whooo: I'll rename it to TSSPrivKey, as that is what it's called in tpm2-openssl

tpm2_pytss/tsskey.py
williamcroberts: > But the format contains the public key, so it's needed for serialization and will be read during deserialization.
> 

Yeah my mistake. I realized this moments after I commented. "ts the TPM private and public blobs from tpm2_load calls.

> I rather see password at load time, so it can be set directly on the ESYS_TR handle, I suspect that those who would use the class directly (not using fromPEM) will do so to call toPEM directly afterwards

OK so the flow (psuedo code) would be:
```Python
pub, priv = Esys_Create(someParent, Template)
key = TSSKey(pub, priv, emptyAuth=True, parent=SomeParent)
pem = key.toPEM()
file.write(pem)
```
or:

```python
pem = file.read()
key = TSSKey.fromPEM(pem)
ctx = Esys_Load(key.pub, key.priv)
# use key
```

I like it. We're just going to want some @property gettr's for getting at the various fields.

whooo: I'll add the property methods.

I was thinking of having load and create methods attached to the class, and the reason for that is how tpm2-openssl handles the parent, it's either by template (if parent is TPM2_RH_OWNER), but it might be a persistent key.

williamcroberts: FYI When I started taking over the Python bindings, the original design had the Esys_TR_setAuth call on the ESYS_TR object itself. So the ESYS_TR object would also have a handle to ectx. Which means you have to be careful that the ESYS context pointer is not freed and closed out from underneath you. So we want to make sure we don't internally keep any esys ctx, just make the caller always pass it. That way the burden of ectx lifecycle is on them.

FYI I also didn't tie it to __del__ because we can't control when or if it's called. In the testing when I had tried this, it would hang occasionally as the ectx wouldn't get closed and the other one would wait while trying to establish a connection with the TPM. Thus you couldn't write code like:

```python
ectx = Esys_Context()
ectx.doSomething()

# This hangs because ectx __del__ was'nt always called
ectx2 = Esys_Context()
ectx2.doSomething()
```

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @property
+    def public(self):
+        return self.public
+
+    @property
+    def emptyAuth(self):
+        if self.emptyAuth:
+            return True
+        return False
```
williamcroberts, line 34: isn't this just a Bool, can't we return it directly?

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,82 @@
...
+            template = parent_rsa_template
+        elif self.parent == lib.TPM2_RH_OWNER and self.public.publicArea.type == TPM2_ALG.ECC:
+            template = parent_ecc_template
+        else:
+            return ectx.TR_GetTpmPublic(self.parent)
+        # generate parent here
+
+    def load(self, ectx, password=None):
+        if not password and self.emptyAuth:
+            raise Exception("no password specified but it is required")
```
williamcroberts, line 59: I think the check is wrong, because if password is not set (ie a string with len 0 or None) and it is the emptyAuth,  that's correct. Ie it is an emptyAuth.

```python
if password and self.emptyAuth:
   raise RuntimeError('Password specified but it has an empty auth value')
```
or
```python
if not password and not self.emptyAuth:
   raise RuntimeError('No password specified but it is required')
```

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,82 @@
...
+        phandle = self._getparent(ectx)
+        handle = ectx.Load(
+            phandle,
+            ESYS_TR_PASSWORD,
+            ESYS_TR_PASSWORD,
+            ESYS_TR_NONE,
+            self.private,
+            self.public,
+        )
+        if password and not self.emptyAuth:
```
williamcroberts, line 69: You already ensured that emptyAuth and password variables are sync'd above, and the SetAuth command handles None, so you can just make this ectx.setAuth(password)

tpm2_pytss/tsskey.py
whooo: it's probably a bool, but it's not certain, so if someone passed emptyAuth=1 to init it will not be a bool.
I can move the check to init, which is probably better

tpm2_pytss/tsskey.py
whooo: yep, it's wrong, I'll fix
the question if we should throw a warning if password is set, but emptyAuth is true, it would be good feedback

tpm2_pytss/tsskey.py
whooo: I guess that depends, so if we pass password and emtyAuth is true it will fail later on, but that might be preferred to pretending that the password was used when it is not

williamcroberts: If emptyAuth is true and pass password, that would mean the check above password was false, which means password is None or the amptyString, the setAuth() would short circuit internally, but setting a password of None or '' IS the emptyAuth, so its NOP.

whooo: after changing the first check, you can pass the password if emptyAuth is true, it currently only check if the password is required and not passed to the method.
I added a warning for that scenario, but it could be changed to a exception

tpm2_pytss/tsskey.py
williamcroberts:
```python
return bool(self.emptyAuth)
```

tpm2_pytss/tsskey.py
williamcroberts: Warning is fine IMO.

tpm2_pytss/tsskey.py
williamcroberts: > after changing the first check, you can pass the password if emptyAuth is true, it currently only check if the password is required and not passed to the method.
> I added a warning for that scenario, but it could be changed to a exception

Yeah I see you changed that to a warning, which is fine by me

whooo:
so it's getting somewhere now, what I see that's left:
Write tests, which might be hard depending on if we should check for tpm2-tss-engine / tpm2-openssl interoperability
Better exceptions (right type, better messages)
Add a create method, so it's easier to create a tssprivkey with the parent templates
Fix the CI job(s)

whooo:
@williamcroberts, this is closer to being usable now, could do a sanity check so I haven't missed something obvious?
I'll need to do some artisan testing with a local tpm2-tss-engine and better unit tests

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,241 @@
...
+            scheme=TPMT_ECC_SCHEME(scheme=TPM2_ALG.NULL),
+            kdf=TPMT_KDF_SCHEME(scheme=TPM2_ALG.NULL),
+        ),
+    ),
+)  # FIXME
+
+loadablekey_oid = ObjectIdentifier("2.23.133.10.1.3")
+
+
+class BooleanOne(Boolean):
```
williamcroberts, line 101: Interesting that the base class ASN1 Boolean didn't have this behavior. What exactly changes between the two?

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,241 @@
...
+    def set(self, value):
+        self._native = bool(value)
+        self.contents = b"\x00" if not value else b"\x01"
+        self._header = None
+        if self._trailer != b"":
+            self._trailer = b""
+
+
+class tssprivkey_der(Sequence):
```
williamcroberts, line 110: Any reason the naming convention changed?
`TSSPrivKeyDer(Sequence):`

williamcroberts (commented):
Overall looks pretty good

tpm2_pytss/tsskey.py
whooo: yeah, my understanding of the DER encoding is that as long as it's not 0 it is true, and asn1crypto encodes True as 0xFF while openssl/tpm2-tss-engine encodes True as 0x01.
While it isn't necessary I think it's good if the same input will lead to the same output as tpm2-tss-engine / tpm2-openssl

tpm2_pytss/tsskey.py
whooo: no reason really, it's only uses for making the (de)serialization easier, so internal use only.
I could always put an underscore before it and perhaps move it to TSSPrivKey

tpm2_pytss/tsskey.py
```diff
@@ -0,0 +1,257 @@
...
+            inPublic=inpub,
+            outsideInfo=TPM2B_DATA(),
+            creationPCR=TPML_PCR_SELECTION(),
+            session1=ESYS_TR.PASSWORD,
+        )
+        return phandle
+
+    def load(self, ectx, password=None):
+        if not password and not self.emptyAuth:
+            raise Exception("no password specified but it is required")
```
williamcroberts, line 199: raise RuntimeError over base class Exception

williamcroberts (changes requested):
I think the last change to pull this in is to get the Exception out of the base class to something more specific.

williamcroberts:
> I think the last change to pull this in is to get the Exception out of the base class to something more specific.

oh and add asn1crypto to setup.cfg:
```diff
diff --git a/setup.cfg b/setup.cfg
index 5c86ec3fa595..95ca351b020a 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ setup_requires =
     cryptography
 install_requires =
     cffi>=1.0.0
+    asn1crypto
 
 [options.extras_require]
 dev =
```

whooo:
> > I think the last change to pull this in is to get the Exception out of the base class to something more specific.
> 
> oh and add asn1crypto to setup.cfg:
> 
> ```diff
> diff --git a/setup.cfg b/setup.cfg
> index 5c86ec3fa595..95ca351b020a 100644
> --- a/setup.cfg
> +++ b/setup.cfg
> @@ -28,6 +28,7 @@ setup_requires =
>      cryptography
>  install_requires =
>      cffi>=1.0.0
> +    asn1crypto
>  
>  [options.extras_require]
>  dev =
> ```

I guess cryptography should be added as well, I'll do a separate PR for that

tpm2_pytss/tsskey.py
whooo: My bad, fixed

williamcroberts:
> ```diff
> setup_requires
> ```

Its in setup_requires, probably should be in install_requires; not 100% sure on those.

williamcroberts (approved):

PR #114, opened 2021-04-20
test: createloaded

williamcroberts:


test/test_esapi.py
```diff
@@ -522,6 +522,36 @@ def test_objectChangeAuth(self):
...
+        creationPCR = TPML_PCR_SELECTION()
+
+        parentHandle, _, _, _, _ = self.ectx.CreatePrimary(
+            ESYS_TR.OWNER, inSensitive, inPublic, outsideInfo, creationPCR
+        )
+
+        templ = TPMT_PUBLIC.parse(
+            alg="rsa2048", objectAttributes=TPMA_OBJECT.DEFAULT_TPM2_TOOLS_CREATE_ATTRS
+        )
+        childInPublic = TPM2B_TEMPLATE(templ.Marshal())
```
williamcroberts, line 543: @whooo thanks for the marshaling code :-p. I wonder if we should implement __bytes__ on TPM_OBJECT to call marshal?

whooo: It would be nice, but it would create some inconsistency with TPM2B_SIMPLE_OBJECT and how that behaves with `__bytes__`

williamcroberts: @whoo I wonder also if we should implement __eq__ as well for types, at least the TPM2B_SIMPLE ones.

whooo: @williamcroberts, heh, I was just thinking about that, I agree it would be useful.
for TPM2B_SIMPLE_OBJECT it should work with bytes, so TPM2B_DIGEST(b"1234") == b"1234" IMO
When it comes to TPM_OBJECT, perhaps doing marshal on both parts and comparing them would work?

williamcroberts: > It would be nice, but it would create some inconsistency with TPM2B_SIMPLE_OBJECT and how that behaves with `__bytes__`

True, but I wonder if that inconsistency matters or is confusing in any way. 

Thinking through this:
- simple: (TPM2B_DIGEST for example) bytes returns just the buffer. (no len)
- compound: (TPMT_PUBLIC for example) bytes returns the serialized data from marshal. (no len prefix)
- complex: (TPM2B_PUBLIC for example) bytes returns the serialized data, which includes len prefix.
  - IIRC these only have one field, so we could call marshal on the sub field that's not size, which would
    make these behave the same as compound.

Issue #104, opened 2021-04-17
Add getName to structures with a nameAlg field


whooo:
It would be useful to get the name NV areas and key objects without a ESAPI context when doing some types of attestation and audit sessions

williamcroberts:
Which, in case you don't know the format, is:

uint16 hash alg + hash(nameAlg, marshalled public)

You can see it in the ESYS code nicely:

KEY: https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-esys/esys_iutil.c#L1507
NV - https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-esys/esys_iutil.c#L1453

whooo:
Yeah, mostly a reminder for myself so I didn't forget about it

PR #115, opened 2021-04-20
implement getName

whooo:
I added TR_GetName mostly to use it in the tests, but I hope it's useful for others as well.

Closes https://github.com/tpm2-software/tpm2-pytss/issues/104


whooo:
I wonder why the style-check is OK at my local setup but not here,

williamcroberts:
> I wonder why the style-check is OK at my local setup but not here,

Black version, you need to 

> I wonder why the style-check is OK at my local setup but not here,

Make sure black version is same:
`black==19.10b0`

williamcroberts (approved):

Issue #117, opened 2021-04-21
Esys_Free on returned pointers


williamcroberts:
Looking through the CFFI documentation, it appears that all the pointers coming back are "owned" and are getting released via whatever ffi.new uses or are getting leaked.  For example consider this:
```python
    def ECC_Parameters(
        self,
        curveID,
        session1=ESYS_TR.NONE,
        session2=ESYS_TR.NONE,
        session3=ESYS_TR.NONE,
    ):

        parameters = ffi.new("TPMS_ALGORITHM_DETAIL_ECC **")
        _chkrc(
            lib.Esys_ECC_Parameters(
                self.ctx, session1, session2, session3, curveID, parameters
            )
        )
        return TPMS_ALGORITHM_DETAIL_ECC(parameters[0])
```

parameters is "owned" by ffi.new and will call it's destructor. The question here is who owns the pointer that paramters[0] refers to? How do we ensure that is free'd? If so how do free it with Esys_Free?

I'm not sure so I posted this question to the CFFI group:
  - https://groups.google.com/g/python-cffi/c/DROgzjrCwE4/m/YzkTduJKBAAJ

PR #120, opened 2021-04-22
ESAPI: free types properly

williamcroberts:
Currently, CData can come from two sources:
1. allocated via ffi.new
2. allocated from a successful call to ESAPI API as returned via the
   double pointer interface.

Item one does not pose an issue, as it will use matching allocator and
deallocator routines.

Item 2 is tricky. Be default (AFAIK[1]), the pointer of the pointer will also be
handled by CFFI's free function. However, that can cause issues where
ESAPI is allocating with one implementation and cffi is deallocating
with another non-abi compatible issue. This issue can be seen with the
introduction of Esys_Free and is notably an issue on Windows.

Thus we need to explicitly, and ONLY for ESAPI allocations, set the
explicit free function. I recieved help from the mailing list[1] for cffi
on how to design this, and tested it by modifying ESAPI to print free
calls. It all seems to be working.

[1] https://groups.google.com/g/python-cffi/c/DROgzjrCwE4/m/vRn55havDAAJ

[2] Worst case I am wrong and CFFI was just leaking that pointer and now
its actually getting free'd.

Fixes: #117

### DIFF
--- a/tpm2_pytss/ESAPI.py
+++ b/tpm2_pytss/ESAPI.py
@@ -9,6 +9,10 @@ from .types import *
 from .utils import _chkrc, TPM2B_pack, TPM2B_unpack
 
 
+def get_ptr(dptr):
+    return ffi.gc(dptr[0], lib.Esys_Free)
+
+
 class ESAPI:
     def __init__(self, tcti=None):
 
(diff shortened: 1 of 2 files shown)


williamcroberts:
@joholl we might need to do the same thing for FAPI?

Issue #122, opened 2021-04-22
Check input params


williamcroberts:
Check params before we call the esapi library so we can give better errors than something like: TypeError: expected integer.

whooo:
Some parameters should be able to be None as well (inPrivate for LoadExternal comes to mind).
It would also be nice if TPM2B_SIMPLE_OBJECT-based parameters could take bytes.

Unless you already started working on it I can throw something together

williamcroberts:
> Some parameters should be able to be None as well (inPrivate for LoadExternal comes to mind).

Yep, I've been doing some of them as I go along. Im working TOP down in ESAPI I just finished EncryptDecrypt2. If you want to start add ESAPI tests starting from the BOTTOM we can meet in the middle. This way we don't have to deal with merge  conflicts constantly and dont dup work.

> It would also be nice if TPM2B_SIMPLE_OBJECT-based parameters could take bytes.

I think this is the case:
 - TPM2B_AUTH: https://github.com/tpm2-software/tpm2-pytss/blob/master/test/test_types.py#L1041
 - TPMS_ECC_POINT: x and y are TPM2B_ECC_PARAMETER (simple) and they get initialized by bytes.

I'd definitely welcome more tests to ensure it all behaves as expected in differing situations.

> 
> Unless you already started working on it I can throw something together

whooo:
> > Some parameters should be able to be None as well (inPrivate for LoadExternal comes to mind).
> 
> Yep, I've been doing some of them as I go along. Im working TOP down in ESAPI I just finished EncryptDecrypt2. If you want to start add ESAPI tests starting from the BOTTOM we can meet in the middle. This way we don't have to deal with merge conflicts constantly and dont dup work.
> 
Sure, sounds good

> > It would also be nice if TPM2B_SIMPLE_OBJECT-based parameters could take bytes.
> 
> I think this is the case:
> 
>     * TPM2B_AUTH: https://github.com/tpm2-software/tpm2-pytss/blob/master/test/test_types.py#L1041
> 
>     * TPMS_ECC_POINT: x and y are TPM2B_ECC_PARAMETER (simple) and they get initialized by bytes.
> 
> 
I was thinking of bytes as arguments, so outsideInfo for Create as an example

williamcroberts:
> > > Some parameters should be able to be None as well (inPrivate for LoadExternal comes to mind).
> > 
> > 
[...]

PR #125, opened 2021-04-26
ESAPI.py: add function to check input parameter types

whooo:
* Allows bytes instead of TPM2B_SIMPLE_OBJECT instances
* Allows None as a parameter when allow_none=True (see inPrivate for LoadExternal)
* Checks for either the correct TPM_OBJECT instance type or the correct ffi cdata instance type

The current tests uses LoadExternal in test_crypto.py and Vendor_TCG_Test which is a bit weird, I can always add more specific tests if it's wanted.

Had to add the get_cdata function to ESAPI.py instead of utils.py as types.py imports utils.py

### DIFF
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -154,6 +154,10 @@ class CryptoTest(TSS2_EsapiTest):
 
         self.ectx.LoadExternal(priv, pub, types.ESYS_TR.RH_NULL)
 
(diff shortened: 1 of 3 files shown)


whooo:
Related to https://github.com/tpm2-software/tpm2-pytss/issues/122

williamcroberts (approved):

PR #127, opened 2021-04-28
Esapi testing: fixes and Certify command

williamcroberts:
William Roberts (3):
      test: remove unused variables from test_Vendor_TCG_Test
      test: fix Vendor_TCG_TEST negative test
      ESAPI: test Certify

### DIFF
--- a/tpm2_pytss/ESAPI.py
+++ b/tpm2_pytss/ESAPI.py
@@ -39,6 +39,13 @@ def get_cdata(value, expected, allow_none=False):
     return value._cdata
 
 
+def check_handle_type(handle, varname):
+    if not isinstance(handle, int):
+        raise TypeError(
+            f"expected {varname} to be type int aka ESYS_TR, got {type(handle)}"
+        )
+
+
 class ESAPI:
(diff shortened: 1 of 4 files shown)


tpm2_pytss/ESAPI.py
```diff
@@ -976,11 +984,20 @@ def Certify(
...
         qualifyingData,
         inScheme,
-        session1=ESYS_TR.NONE,
-        session2=ESYS_TR.NONE,
+        session1=ESYS_TR.PASSWORD,
+        session2=ESYS_TR.PASSWORD,
         session3=ESYS_TR.NONE,
     ):
 
+        check_handle_type(objectHandle, "objectHandle")
```
williamcroberts, line 992: @whooo just a few thoughts before I head out on a few days of vacation (back May 3):
  - We probably want to type check the handles, so I added this routine.
  - The type check of other data, the get_cdata routine you added, we should probably pass the variable name so TypeErrors say explicitly that this argument failed. This way users get explicit feedback that should help avoid ambiguity.

Thanks again for your help along the way, much appreciated.

PR #136, opened 2021-05-12
tests for Clock functions

whooo:


williamcroberts:
Merged manually as:
d161e164815f test_esapi.py: add ClockRateAdjust test
19ef86d977ac test_esapi.py: add ClockSet test
77752a3eb512 test_esapi.py: add ReadClock test

PR #142, opened 2021-05-17
Add support for pure DER encoded keys and certificates

whooo:
With this most common formats/encodings are supported directly, not just PEM encoded public keys.

I also added cryptography to install_requirements.

### DIFF
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -8,7 +8,7 @@ import unittest
 
 from tpm2_pytss import *
 from .TSS2_BaseTest import TSS2_EsapiTest
-
+from base64 import b64decode
 
 rsa_private_key = b"""
 -----BEGIN RSA PRIVATE KEY-----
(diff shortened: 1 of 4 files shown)


test/test_crypto.py
```diff
@@ -208,3 +262,50 @@ def test_nv_getname(self):
         ename = self.ectx.TR_GetName(handle)
 
         self.assertEqual(ename.name, oname.name)
+
+    def test_public_from_pem_rsa_pem_cert(self):
+        pub = TPMT_PUBLIC()
+        crypto.public_from_pem(rsa_cert, pub)
```
williamcroberts, line 268: maybe refactor the function name to public_from_cert?

whooo: I'll rename it to public_from_encoding so the name is generic

williamcroberts (approved):

PR #147, opened 2021-05-18
[RFC] Quote thoughts

williamcroberts:
@whooo and @AndreasFuchsSIT, the last commit makes some changes to quote API. I would like to get thoughts and feedback from both of you on it. I have this thought applying to a lot of the API, like CreatePrimary for PCR Selections and other data that's usually empty.

### DIFF
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -1874,6 +1874,117 @@ class TestEsys(TSS2_EsapiTest):
         with self.assertRaises(TypeError):
             self.ectx.Rewrap(primary2, primary1, duplicate, keyName, TPMT_PUBLIC())
 
+    def test_Import(self):
+
+        sym = TPMT_SYM_DEF(
+            algorithm=TPM2_ALG.XOR,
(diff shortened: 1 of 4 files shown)


whooo:
I'm OK with changing the API (not having to pass ousideInfo and creationPCR to Create would be nice...) and it's already done for part of the API (at least NV_Write and NV_Read).

But when it comes to Quote and qualifyingData I think it still should be required to be provided by the caller as it's the only part stopping the replay of a previous quote.

williamcroberts:
> I'm OK with changing the API (not having to pass ousideInfo and creationPCR to Create would be nice...) and it's already done for part of the API (at least NV_Write and NV_Read).
> 
> But when it comes to Quote and qualifyingData I think it still should be required to be provided by the caller as it's the only part stopping the replay of a previous quote.

Yeah totally agree on QualifyingData. Right after I posted this, I was like I am going to make that mandatory and require folks to choose their destiny. So I guess then we generally agree, so @whooo as you go through things, feel free to make API changes if you desire. Now is the time :-p

williamcroberts:
Changes brought to: https://github.com/tpm2-software/tpm2-pytss/pull/148

PR #148, opened 2021-05-19
Esapi testing (Quote) and API changes to CreatePrimary and Create Plus lots of checks/tests added

williamcroberts:
ESAPI: update CreatePrimary API

Make CreatePrimary interface:
  - not need PCR data, default empty
  - not need creation data, default empty
    - creation data since it's a pcr selection supports pcr selection
      strings.
  - not need hierarchy, default owner
  - support tpm2-tools like initializer strings for TPM2B_PUBLIC
    - Defualt name alg SHA256
    - Default attribute set

It will stillr require the sensitive portion so folks have to explicitly
opt in to no passwords.

Additional Notes:
I did initially code this in a way where the get_cdata, on getting passed a string, if the current class object didn't have a parse routine, would look at the subfields and if only one subfield other than size existed, instantiate that, get the cdata and then use than call that subclass's parse method. While more generic, it was a little more code, more complex to read, and the current usecase is only for TPM2B_PUBLIC. Thus, writing generic code for a one off case seemed more complicated than it needed to be. So I ended up just adding parse to TPM2B_PUBLIC.

### DIFF
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -163,7 +163,31 @@ class CryptoTest(TSS2_EsapiTest):
 
         priv = types.TPM2B_SENSITIVE.fromPEM(rsa_private_key)
 
-        self.ectx.LoadExternal(priv, pub, types.ESYS_TR.RH_NULL)
+        # test without Hierarchy
+        handle = self.ectx.LoadExternal(priv, pub)
+        self.assertNotEqual(handle, 0)
+
+        # negative test
+        with self.assertRaises(TypeError):
+            self.ectx.LoadExternal(TPM2B_PUBLIC(), pub)
+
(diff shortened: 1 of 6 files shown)


williamcroberts:
@whooo FYI what we discussed about CreatePrimary, if you could look it over. Ill go through and do the same thing with Create next, once we iron out anything you don't like here. Thanks as always.

williamcroberts:
I'm going to merge this tomorrow if I don't get any feedback as I have 39 commits over master. Large out of tree series i'd like to merge.

whooo:
> I'm going to merge this tomorrow if I don't get any feedback as I have 39 commits over master. Large out of tree series i'd like to merge.

Looks good to me

PR #150, opened 2021-05-22
Misc tests

whooo:
Tests for the functions under "Miscellaneous Management Functions"

### DIFF
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -2867,6 +2867,24 @@ class TestEsys(TSS2_EsapiTest):
         with self.assertRaises(TypeError):
             self.ectx.PP_Commands(TPML_CC(), TPML_CC(), authHandle="platform")
 
+    def test_SetAlgorithmSet(self):
+        self.ectx.SetAlgorithmSet(0)
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet([1, 2, 3])
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet(session2=set(3, 2, 1))
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet(session1=set(4, 3, 2))
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet(session3=set(5, 4, 3))
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet(authHandle=None)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -2843,6 +2843,30 @@ class TestEsys(TSS2_EsapiTest):
                 signHandle, session, b"1234", session3=12.723
             )
 
+    def test_PP_Commands(self):
+        with self.assertRaises(TSS2_Exception) as e:
+            self.ectx.PP_Commands(TPML_CC(), TPML_CC(), session1=ESYS_TR.PASSWORD)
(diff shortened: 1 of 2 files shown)


tpm2_pytss/ESAPI.py
```diff
@@ -2431,21 +2431,31 @@ def PP_Commands(
...
 
+        check_handle_type(auth, "auth")
+        setList_cdata = get_cdata(setList, TPML_CC, "setList")
+        clearList_cdata = get_cdata(clearList, TPML_CC, "clearList")
         _chkrc(
             lib.Esys_PP_Commands(
-                self.ctx, auth, session1, session2, session3, setList, clearList
+                self.ctx,
+                auth,
+                session1,
```
williamcroberts, line 2441: Why not default session1 to password session?

test/test_esapi.py
```diff
@@ -2843,6 +2843,17 @@ def test_GetSessionAuditDigest(self):
                 signHandle, session, b"1234", session3=12.723
             )
 
+    def test_PP_Commands(self):
+        with self.assertRaises(TSS2_Exception) as e:
+            self.ectx.PP_Commands(
+                ESYS_TR.PLATFORM, TPML_CC(), TPML_CC(), session1=ESYS_TR.PASSWORD
+            )
+        self.assertEqual(e.exception.error, TPM2_RC.PP)
+        self.assertEqual(e.exception.session, 1)
```
williamcroberts, line 2852: negative test all inputs.

tpm2_pytss/ESAPI.py
whooo: PP_Commands seems to require a policy session as you have to assert physical presence

test/test_esapi.py
whooo: do you mean "bad" inputs, such as wrong type?

tpm2_pytss/ESAPI.py
williamcroberts: Duh, I misread the test, that's a negative test:
```python
        with self.assertRaises(TSS2_Exception) as e:
```

test/test_esapi.py
williamcroberts: Yes

whooo: Added some tests to check for TypeError when passing bad input

whooo (commented):
Added some negative tests for both commands and changed auth(Handle) to use the default value of ESYS_TR.PLATFORM as bot commands requires platform authoziation anyway

test/test_esapi.py
```diff
@@ -2843,6 +2843,30 @@ def test_GetSessionAuditDigest(self):
...
+            self.ectx.PP_Commands(TPML_CC(), None, session1=b"0xF1F1")
+
+    def test_SetAlgorithmSet(self):
+        self.ectx.SetAlgorithmSet(0)
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet([1, 2, 3])
+
+        with self.assertRaises(TypeError):
+            self.ectx.SetAlgorithmSet(session2=set(3, 2, 1))
```
williamcroberts, line 2868: Test each and every input as tedious as it is:
1. This is to ensure we get all the TypeError behavior right in the API.
2. This also helps bump test coverage, because we're testing the exception paths.

Look at the tests, like test_CreatePrimary:
  - https://github.com/tpm2-software/tpm2-pytss/blob/master/test/test_esapi.py#L82

for an example.

whooo: Sure, it should be done now and I'll remember that for the next set of ESAPI tests

williamcroberts (approved):

PR #154, opened 2021-05-24
add python MakeCredential implementation

whooo:
This will add a MakeCredential implementation without the need for a TPM (simulator) and lay the foundation for the duplicate protocol


whooo:
@williamcroberts, it works for me now, but if you have time to give it a look/try please do.

I'll do some clean up of the code and see if I can add some more tests, but the foundation is there now at least

williamcroberts (approved):

Issue #159, opened 2021-06-03
Expose fapi_info


IshFx:
Hello,

About the TODO here : https://github.com/tpm2-software/tpm2-pytss/blob/07a92e9fa75548ea102ce90b3b6182093b3f7a73/tpm2_pytss/FAPI.py#L157

How about simply exposing the FapiInfo instance ?

- Renaming `info` method to `get_info` because its not a property and it calls a C++ function GetInfo under the hood.
- Adding a property `info` that returns an instance of FapiInfo using the `get_info()` string data.

```py
def get_info(self) -> str:  # TODO other type? json?
    """Get Fapi information, containing library info, TPM capabilities and more.

    Raises:
        TSS2_Exception: If Fapi returned an error code.

    Returns:
        str: JSON-encoded info string.
    """
    info = ffi.new("char **")
    ret = lib.Fapi_GetInfo(self.ctx, info)
    if ret == lib.TPM2_RC_SUCCESS:
        result = ffi.string(info[0]).decode(self.encoding)
        lib.Fapi_Free(info[0])
        return result
    raise TSS2_Exception(ret)

@property
def info(self):
    info_data = json.loads(self.get_info())
    return FapiInfo(info_data)
```

This way there is no need to have multiple properties (like `version`, `config`, etc).
They would be accessible through `info` property using dot notation.

I can submit a PR if requested.

Thanks in advance !

williamcroberts:
what not in FAPI `__init__` call the lib.FAPI_GetInfo and cache the value and have the info be a property that returns the cached info. I'm assuming that getinfo doesn't return non-static values.

However, the spec does say, "Fapi_GetInfo() returns a UTF-8 string identifying the version of the FAPI, the TPM, configurations and other relevant
information in a human readable format. The concrete content of this string is implementation specific."

I'm okay with making this a property, but I am not sure if we should interpret the string as JSON, as it might not be.

williamcroberts:
@joholl thoughts?

williamcroberts:
@AndreasFuchsSIT thoughts as well?

IshFx:
> what not in FAPI __init__ call the lib.FAPI_GetInfo and cache the value and have the info be a property that returns the cached info. I'm assuming that getinfo doesn't return non-static values.

I don't know if it's a good idea to call the GetInfo at instantiation because this would directly send a "TPM2_GetCapability" command to the TPM. Interacting with the TPM at instantiation might not be wanted by the user.

[...]

Issue #54, opened 2021-03-04
Add FAPI


williamcroberts:
Add the feature api and test

williamcroberts:
@joholl are you ready for FAPI to ship?

joholl:
Sorry I did not push in quite some time. I will push tomorrow again. I've basically implemented everything but policy authorized by now :)

williamcroberts:
> Sorry I did not push in quite some time. I will push tomorrow again. I've basically implemented everything but policy authorized by now :)

No worries, just close this ticket when you're ready so I know when I can release.

joholl:
With #160, technically all Fapi calls are implemented. That is, if I did not forgot any.

Open issues:
* currently, we test against the default FAPI profile. If someone is using a custom one, the tests will fail. Instead, we should ship the testing profile(s) with the tests
* since we only test with the ECC profile atm, the encrypt and decrypt calls are _not tested_
* since the simulator does not ship a platform cert by default, the get_platform_certificate() call is practically _not tested_

We probably want to fix these before the release.

williamcroberts:
> With #160, technically all Fapi calls are implemented. That is, if I did not forgot any.
> 
> Open issues:
> 
> * currently, we test against the default FAPI profile. If someone is using a custom one, the tests will fail. Instead, we should ship the testing profile(s) with the tests
> * since we only test with the ECC profile atm, the encrypt and decrypt calls are _not tested_
> * since the simulator does not ship a platform cert by default, the get_platform_certificate() call is practically _not tested_
> 
> We probably want to fix these before the release.

FYI we can mock the lib layer for things that are practically difficult to do in integration tests if that helps you get some testing around these?

williamcroberts:
@joholl are we ready to ship the FAPI portion yet?

joholl:
Almost. There are two things left to do:

* I'd like to add one parameter to the API, PR will follow
* We should agree on how to handle acronyms with regards to naming convention... (FapiConfig vs FAPI). This applies to classes and file names (which are kind of also part of the API). PEP-8 recommends all uppercase, I personally are indifferent. We should make it consistent, though. What do you think?

williamcroberts:
> Almost. There are two things left to do:
> 
> * I'd like to add one parameter to the API, PR will follow

Ok

[...]

PR #160, opened 2021-06-03
fapi: add more api calls

joholl:
Add every fapi api call besides policy authorize.

#54


joholl:
This PR is gotten a bit long, I hope that's ok.
 * add the last api call `set_policy_action_callback()`
* rename `fapi.info()` to `fapi.get_info()` and have it return the info as string (Fixes #159 for now, details of parsing remain TBD)
 * add workaround for tpm2-software/tpm2-tss#2091
 * use `_get_ptr()` and `_chkrc()` instead of manually freeing memory and checking return codes (Fixes  #165)

Sorry that this took longer than expected, I needed to debug quite some TSS. I'll give some status information in #54.

williamcroberts (approved):

Issue #49, opened 2020-12-04
Feature Request: set TSS log level per logging module


joholl:
Unfortunately, there seems to be no mechanism to set the TSS log level at runtime besides via the env variable `TSS2_LOG`, e.g. `TSS2_LOG=all+ERROR,fapi+TRACE`, as documented [here](https://github.com/tpm2-software/tpm2-tss/blob/master/doc/logging.md). What do you think of having a function setting this variable, maybe from a dict?

Example:

```python
log_levels = { "all": "error", "tcti": "trace"}
log_levels_str = ",".join(f"{k}+{v}" for k, v in log_levels.items())  # 'all+error,tcti+trace'
os.environ["TSS2_LOG"] = log_levels_str
```

We might want to use enums instead of strings(?). In case you wonder, modules and log levels are parsed **case insensitive**.

Complete list of TSS log levels (as defined [here](https://github.com/tpm2-software/tpm2-tss/blob/master/src/util/log.h#L39)):
* none
* (unused)  [note: defined with parens]
* ERROR
* WARNING
* info
* debug
* trace

Complete list of TSS log modules (grepped):
* fapi
* esys
* esys_crypto
* sys
* marshal
* tcti
* log
* test
* all  [note: covers all modules]

pdxjohnny:
Sweet. I'm doing a major re-write at the moment. I'll make sure to get this in there. Thanks!

williamcroberts:
We don;t have a log in the rewrite to CFFI yet, but we will add it.

joholl:
Nice, I did some preliminary work on this. That includes a PR for redirecting logging to a file (which could be a pipe) on TSS side,  see tpm2-software/tpm2-tss#1976

williamcroberts:
I don't think this is going to make 1.0 and we need to finalize the tss C code approach. We can't do per-ctx logging because much of the submodules don't have context parameters, like libmu.

See:
  - https://github.com/tpm2-software/tpm2-tss/pull/2087

PR #161, opened 2021-06-06
utils: add functions to get and set logging levels for tpm2-tss

whooo:
Related to https://github.com/tpm2-software/tpm2-pytss/issues/49

I skipped support for TSS2_LOGFILE for now and due to how the logging works in libtss2 the loglevel can't be changed for a module as soon as that module have called a logging function

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -1,0 +1,47 @@
+#!/usr/bin/python3 -u
+"""
+SPDX-License-Identifier: BSD-2
+"""
+import os
+import unittest
+
+from tpm2_pytss.utils import *
+
+
+class TypesTest(unittest.TestCase):
+    def test_get_logging(self):
+        if "TSS2_LOG" in os.environ:
+            del os.environ["TSS2_LOG"]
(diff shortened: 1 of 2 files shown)


joholl:
Cool, this is much needed!

I'm not sure if I forgot the `fapijson` [log module](https://github.com/tpm2-software/tpm2-tss/search?q=fapijson) or if it was added since then. We should add that.

whooo:
@joholl, just pushed an update with fapijson and removed the "test" module as it only seems to be used for the tests

joholl:
Hm, I thought about this a bit. I feel like a pythonic module should not write to stderr, at all. Instead, it should hand log records to the python standard lib [logging module](https://docs.python.org/3/howto/logging.html). This has many advantages: the user can have one log file for everything (including application-level logging), use the python log handlers and log levels, have one log format etc.

Have a look at draft #164.

whooo:
if something like https://github.com/tpm2-software/tpm2-tss/pull/2087 is implemented I don't think this is needed

Issue #55, opened 2021-03-04
Add TCTI


williamcroberts:
Add the TCTI API

joholl:
Currently, the ESAPI `__init__()` function expects a `TctiLdr` instance (or to be more specific, a python object with a`.ctx` attribute. This `.ctx` is the actual TCTI cdata object.

https://github.com/tpm2-software/tpm2-pytss/blob/07a92e9fa75548ea102ce90b3b6182093b3f7a73/tpm2_pytss/ESAPI.py#L64-L71

We should expect the TCTI cdata context directly. This way, the user can not only pass TctiLdr instances, but also TCTI instances.

Because I could not pass a TCTI object directly, I needed to use an awkward workaround in the FAPI tests in #54, for now.

```python
    @pytest.fixture
        def esys(self):
            # TODO ESAPI should accept either a cdata obj or a dedicated TCTI class, not TctiLdr!
            tcti = SimpleNamespace()
            tcti.ctx = self.fapi.tcti
            with ESAPI(tcti=tcti) as esys:
                yield esys
``

whooo:
I suspect this issue can be closed now

PR #176, opened 2021-06-19
add TCTI class for the TCTI API

whooo:
With this a clear API for passing a TCTI context is possible.
As well as providing access to the low level TCTI API

Related to https://github.com/tpm2-software/tpm2-pytss/issues/55

I'm not able to test get_poll_handles for an unknown reason (mssim should be able to handle it), might be something with how the bindings are being generated and built.
But I'm not so sure that it's useful at all, I just added it for completeness.

The same issue with testing exists for make_sticky as mssim doesn't support it either, but I think it's useful if you want use for example the linux kernel methods for dealing with TPM keys.

If this is merged I'll make another PR so that TctiLdr return an (sub-)instance off the TCTI class and that ESAPI and FAPI will expect the tcti_context property

### DIFF
--- a/test/test_fapi.py
+++ b/test/test_fapi.py
@@ -5,7 +5,6 @@ SPDX-License-Identifier: BSD-2
 import binascii
 import random
 import string
-from types import SimpleNamespace
 
 import pkgconfig
 
@@ -71,10 +70,7 @@ def init_fapi(request, fapi):
 class TestFapi:
     @pytest.fixture
     def esys(self):
-        # TODO ESAPI should accept either a cdata obj or a dedicated TCTI class, not TctiLdr!
-        tcti = SimpleNamespace()
-        tcti.ctx = self.fapi.tcti
-        with ESAPI(tcti=tcti) as esys:
+        with ESAPI(tcti=self.fapi.tcti) as esys:
             yield esys
 
     @pytest.fixture
--- a/test/test_tcti.py
+++ b/test/test_tcti.py
@@ -1,0 +1,66 @@
+#!/usr/bin/python3 -u
+"""
+SPDX-License-Identifier: BSD-2
+"""
+
+import unittest
+
+from tpm2_pytss import *
+from .TSS2_BaseTest import TSS2_EsapiTest
+
+
+class TestTCTI(TSS2_EsapiTest):
+    def test_init(self):
+        self.assertEqual(self.tcti.version, 2)
+        self.assertGreater(self.tcti.magic, 0)
+
+        v1ctx = ffi.cast("TSS2_TCTI_CONTEXT_COMMON_V1 *", self.tcti._ctx)
+        v1ctx.version = 1
+
+        tcti = TCTI(self.tcti._ctx)
+        self.assertEqual(tcti.version, 1)
+        self.assertEqual(tcti._v2, None)
+
+    def test_transmit_receive(self):
+        startup = b"\x80\x01\x00\x00\x00\x0C\x00\x00\x01\x44\x00\x00"
+        self.tcti.transmit(startup)
+
+        resp = self.tcti.receive()
+        self.assertEqual(resp, b"\x80\x01\x00\x00\x00\n\x00\x00\x01\x00")
+
+    def test_finalize(self):
+        tcti = TCTI(self.tcti._ctx)
+        tcti.finalize()
+
+    def test_cancel(self):
+        if getattr(self.tcti, "name", "") == "swtpm":
+            self.skipTest("cancel supported by swtpm")
+
+        startup = b"\x80\x01\x00\x00\x00\x0C\x00\x00\x01\x44\x00\x00"
+        self.tcti.transmit(startup)
+        self.tcti.cancel()
+
+    def test_get_poll_handles(self):
+        tcti_name = getattr(self.tcti, "name", "")
+        try:
+            handles = self.tcti.get_poll_handles()
+        except TSS2_Exception as e:
+            if e.rc != lib.TSS2_TCTI_RC_NOT_IMPLEMENTED:
+                raise e
+            else:
+                self.skipTest(f"get_poll_handles not supported by {tcti_name}")
+
+    def test_set_locality(self):
+        self.tcti.set_locality(TPMA_LOCALITY.TWO)
+
+    def test_make_sticky(self):
+        tcti_name = getattr(self.tcti, "name", "")
+        if tcti_name in ("swtpm", "mssim"):
+            self.skipTest(f"make_sticky not supported by {tcti_name}")
+        raise Exception(self.tcti.name)
+        self.tcti.make_sticky(0, 0)
+
+        tcti._v2 = None
+        with self.assertRaises(RuntimeError) as e:
+            self.tcti.make_sticky(0, 0)
+        self.assertEqual(str(e.exception), "unsupported by TCTI API version")
--- a/tpm2_pytss/ESAPI.py
+++ b/tpm2_pytss/ESAPI.py
@@ -7,6 +7,7 @@ from ._libtpm2_pytss import lib
 from .types import *
 
 from .utils import _chkrc, TPM2B_pack
+from .TCTI import TCTI
 
 
 def get_ptr(dptr):
@@ -72,7 +73,8 @@ def check_friendly_int(friendly, varname, clazz):
 class ESAPI:
     def __init__(self, tcti=None):
 
-        tctx = ffi.NULL if tcti is None else tcti.ctx
+        self._tcti = tcti
+        tctx = ffi.NULL if tcti is None else tcti._tcti_context
 
         self.ctx_pp = ffi.new("ESYS_CONTEXT **")
         _chkrc(lib.Esys_Initialize(self.ctx_pp, tctx, ffi.NULL))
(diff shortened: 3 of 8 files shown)


williamcroberts:
Shouldn't TCTILdr return this class and shouldn't ESAPI init only take a TCTI object. I think we could update the base class for tests to use this new class.

whooo:
> Shouldn't TCTILdr return this class and shouldn't ESAPI init only take a TCTI object. I think we could update the base class for tests to use this new class.

Yes, that's my general idea, I just want some feedback before I dig into porting all users of the TCTI context to this and if get_poll_handles is useful at all (I suspect it's not)
Regarding TctiLdr, I was thinking of using a subclass so the name and conf can be part of the class, useful for skipping some tests if needed (the cancel test when using swtpm for example)

williamcroberts:
> Yes, that's my general idea, I just want some feedback before I dig into porting all users of the TCTI context to this and if  get_poll_handles is useful at all (I suspect it's not)

For non-threading non-blocking code it is, and the fd's are usable by the os methods. Looks good thus far, no major issues from my end.

> Regarding TctiLdr, I was thinking of using a subclass so the name and conf can be part of the class, useful for skipping some tests if needed (the cancel test when using swtpm for example)

I don't think im picking up what you're putting down here fully. Code example?

whooo:
> > Regarding TctiLdr, I was thinking of using a subclass so the name and conf can be part of the class, useful for skipping some tests if needed (the cancel test when using swtpm for example)
> 
> I don't think im picking up what you're putting down here fully. Code example?

```class TctiLdrCtx(TCTI):
        def __init__(self, ctx, name, conf):
                super().__init__(self, ctx)
                self.name = name
                self.conf = conf

...
def test(self):
        if getattr(self.tcti, "name", "") == "swtpm":
                return

```

whooo:
@joholl, could you have a look at the change for the FAPI tests? I don't know if there is a better way to handle it

test/test_fapi.py
```diff
@@ -72,8 +72,7 @@ class TestFapi:
     @pytest.fixture
     def esys(self):
         # TODO ESAPI should accept either a cdata obj or a dedicated TCTI class, not TctiLdr!
-        tcti = SimpleNamespace()
-        tcti.ctx = self.fapi.tcti
+        tcti = TCTI(self.fapi.tcti)
```
williamcroberts, line 75: shouldn't FAPI's property function `tcti` return the TCTI class?

test/test_tcti.py
```diff
@@ -0,0 +1,64 @@
...
+
+        resp = self.tcti.receive()
+        self.assertEqual(resp, b"\x80\x01\x00\x00\x00\n\x00\x00\x01\x00")
+
+    def test_finalize(self):
+        tcti = TCTI(self.tcti._ctx)
+        tcti.finalize()
+
+    def test_cancel(self):
+        if getattr(self.tcti, "name", "") == "swtpm":
```
williamcroberts, line 36: I think we want to use:
```python
self.skipTest()
```

test/test_tcti.py
```diff
@@ -0,0 +1,64 @@
...
+        startup = b"\x80\x01\x00\x00\x00\x0C\x00\x00\x01\x44\x00\x00"
+        self.tcti.transmit(startup)
+        self.tcti.cancel()
+
+    def test_get_poll_handles(self):
+        try:
+            handles = self.tcti.get_poll_handles()
+        except TSS2_Exception as e:
+            if e.rc != lib.TSS2_TCTI_RC_NOT_IMPLEMENTED:
+                raise e
```
williamcroberts, line 49:
``` python
else:
    self.skipTest
```

test/test_tcti.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def test_set_locality(self):
+        self.tcti.set_locality(TPMA_LOCALITY.TWO)
+
+    def test_make_sticky(self):
+        if getattr(self.tcti, "name", "") in ("swtpm", "mssim"):
+            # neither swtpm nor the mssim TCTI modules doesn't support makeSticky
+            return
```
williamcroberts, line 57: skip

tpm2_pytss/ESAPI.py
```diff
@@ -72,7 +72,8 @@ def check_friendly_int(friendly, varname, clazz):
 class ESAPI:
     def __init__(self, tcti=None):
 
-        tctx = ffi.NULL if tcti is None else tcti.ctx
+        self._tcti = tcti
+        tctx = ffi.NULL if tcti is None else tcti.tcti_context
```
williamcroberts, line 76: does a tcti have a non `tcti_context`? wouldn't juct tcti.context be more succinct without losing information?

whooo: Which tcti do you mean? self._tcti is for (a later) GetTcti method

williamcroberts: Ahh I misread it, but the `tcti.tcti_context` is the FFI pointer, right? If so, we would want that to be behind a "private" attribute like `tcti._context` or `tcti._ctx`. We want to make it clear that folks shouldn't be using the FFI values by convention.

Unrelated to what you're doing, we'd probably want to make this change in ESAPI as well, I see it's `self.ctx` (sorry my mistake there, ill fix it before release #178.)

whooo: tcti_context is the FFI pointer, exactly, and it's a property of the TCTI class, self._tcti is later getting the TCTI classed passed to init (so via GetTcti and/or a property of the ESAPI class).

I'll add GetTcti and the tcti property to ESAPI in this PR

williamcroberts: > tcti_context is the FFI pointer, exactly, and it's a property of the TCTI class, self._tcti is later getting the TCTI classed passed to init (so via GetTcti and/or a property of the ESAPI class).

Can we make the property an underscore property as the FFI types aren't part of the public API.

> 
> I'll add GetTcti and the tcti property to ESAPI in this PR

tpm2_pytss/TCTI.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, ctx):
+        self._v1 = ffi.cast("TSS2_TCTI_CONTEXT_COMMON_V1 *", ctx)
+        if self._v1.version == 2:
+            self._v2 = ffi.cast("TSS2_TCTI_CONTEXT_COMMON_V2 *", ctx)
+        else:
+            self._v2 = None
+        self._ctx = ctx
+
+    @property
+    def tcti_context(self):
```
williamcroberts, line 20: I don't want public API that returns an FFI pointer, so I want this one to go away. Where you call it, just call ._ctx and consider it "Package Scope".

whooo: Ah, I see, my general idea was to enable flexibility on how applications configure/use TCTI and by only expecting a certain property to be available it provides some room to do non-standard things.
I can always rename the property to _tcti_context instead, to make it clear that it's not expected to by used outside of the package

test/test_fapi.py
joholl: Yes. The TCTI should make the underlying `cdata` object available anyway.

tpm2_pytss/TCTI.py
joholl: @williamcroberts Maybe I understood this incorrectly, so please help me out here.

I think we **need** the TCTI class to expose the underlying `cdata` object. The FAPI returns only a TCTI object. If there was no way to get the `cdata` object from the TCTI class, a user couldn't get the `cdata` TCTI from FAPI.

williamcroberts: The class itself will need to make the context as the cdata pointer available for calls. However, in Python, anything not prefixed with an underscore is considered public API and free-to-call by end users. Thus, I don't want end users directly asking for the cdata pointer, so all I am asking for here is to change.

williamcroberts: @joholl The class itself will need to make the context as the cdata pointer available for calls. However, in Python, anything not prefixed with an underscore is considered public API and free-to-call by end users. Thus, I don't want end users directly asking for the cdata pointer, so all I am asking for here is to change.

williamcroberts: > Ah, I see, my general idea was to enable flexibility on how applications configure/use TCTI and by only expecting a certain property to be available it provides some room to do non-standard things.
> I can always rename the property to _tcti_context instead, to make it clear that it's not expected to by used outside of the package

I would rather not build this on by default. For know folks that need something can just call the underscore variant, we can keep that stable enough. I would like to hear the usecases and fix the API. I look at this as being similar to all the work we did wrapping types.

williamcroberts (approved):
@whooo LGTM, you ready for merge?

whooo:
> @whooo LGTM, you ready for merge?

I'm ready for merge

williamcroberts:
merged manually due to conflicts: 87e379e61513..8219c9d0f781

tpm2_pytss/TCTI.py
joholl: Hm, I have no strong opinion on this. Calling internal-only properties is considered bad practice and shouldn't be done anyway. Currently I cannot think of a use case where a user needs the TCTI cdata. Internal it is, then.

PR #180, opened 2021-06-23
Hierarchy command test

whooo:
Add tests for the missing command from hierarchy control


williamcroberts:
merged out of band to correct conflicts. HEAD as 87e379e61513

Issue #163, opened 2021-06-07
Converge Naming


joholl:
Moved from #159.

> @joholl and I also diverged in ESAPI vs FAPI function naming. I went camelcase and @joholl went underscore. I'd also like to converge before release so the code base looks more consistent.

I also noticed that. There is an argument to be made for camel case: consistency with the spec. However, I'd rather go for _pythonic_-ness, that is adherence to the [python naming conventions](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), i.e. snake case.

Also, we should align ESAPI vs Fapi. According to PEP 8, [python classes are camel case](https://www.python.org/dev/peps/pep-0008/#class-names).

@williamcroberts do you have a preference?

williamcroberts:
- [ ] ESAPI PR #190
 - [ ] TCTI
 - [ ] TCTILdr
 - [ ] TSS2_Exception 
 - [ ] types

PR #182, opened 2021-06-25
Fix imports and add isort

joholl:
We have wildcard imports (from ... import *`). That is bad practice.
Additionally, the imports are not sorted alphabetically, currently.

Fix imports and add isort to linters. I also added a linters script to call before a commit for convenience:

```bash
$ .ci/linters.sh
```

Eventually, I'd like to add `pylint`, a de-facto standard and maybe `mypy` and `pydocstyle`.


williamcroberts (dismissed):
Generally I try and leave large sweeping formatting changes till the end, so folks aren't battling  oodles of conflicts over more productive tasks. @whooo are you good with these now? I'm fine with them, they merge cleanly in my tree.

williamcroberts:
tests are failing, not sure why.

joholl:
Oh ok, I don't want to produce annoying conflicts. This is an attempt to battle technical dept and one step on the way to fix #163 eventually (because we need a linter for that and at least pylint floods stderr with import issues).

If you prefer that, I'll postpone big linter fixes.

whooo:
I don't generally have anything against this, but when it comes to types.py and the tests I rather see a wildcard import, at least for test_esapi.py, as the import list will probably continue to grow and being able to easily throw a test together is good, at least for my workflow :-)

I think keeping __init__.py would be nice as well, doing something like ```from tpm2_pytss import ESAPI, TPM2B_DIGEST``` is useful. One thing I see the python bindings being really good for is easy prototyping and just testing something out so just being able to do ```from tpm2_pytss import *``` and then continue with your code would provide a good experience

test/test_esapi.py
```diff
@@ -3,9 +3,83 @@
...
+    TPMT_PUBLIC_PARMS,
+    TPMT_RSA_DECRYPT,
+    TPMT_SIG_SCHEME,
+    TPMT_SIGNATURE,
+    TPMT_SYM_DEF,
+    TPMT_SYM_DEF_OBJECT,
+    TPMT_TK_CREATION,
+    TPMT_TK_HASHCHECK,
+    TPMT_TK_VERIFIED,
+)
```
whooo, line 57: These should be imported from tpm2_pytss.types

test/test_types.py
```diff
@@ -6,7 +6,38 @@
...
+    TPML_ALG,
+    TPML_DIGEST_VALUES,
+    TPML_PCR_SELECTION,
+    TPMS_ECC_POINT,
+    TPMS_PCR_SELECTION,
+    TPMS_SENSITIVE_CREATE,
+    TPMT_HA,
+    TPMT_PUBLIC,
+    TPMU_HA,
+    ffi,
```
whooo, line 39: ffi should come from _libtpm2_pytss

williamcroberts:
Ahh I just noticed that __init__.py is getting dropped. Yeah we want to keep that IIUC, but this area of python is my least knowledgeable area. At the end of the day, I want the most simple example to be `from tpm_pytss import *`, and it just works. While import * is frowned upon, once an import list gets really large, I think its simpler then to have to keep adding every little thing and minimizes the size of the python file. No one wants a a 50+ line import statement.

test/test_esapi.py
joholl: You're right, my bad

test/test_types.py
joholl: Yup

joholl:
Yes, I threw out all asterisk imports in the pytss code. For rapid coding we (and the user) can still use it, though. Removing the imports from `__init__.py` just means that you have to import the names via dot notation:

```python
from tpm2_pytss.types import *
from tpm2_pytss.ESAPI import *
# ...
```

I can see that the imports in our own code are long-ish now (on my side, my pycharm hides them for me). I could easily change the formatting so that we do not have one import per line.

I understand that is can be slightly more effort to use the dot notation here, but I feel it adds clarity and it seems to be the python way to do things. There are probably good reasons but, to be frank, I am not the most experienced python expert either ;)

whooo:
> Yes, I threw out all asterisk imports in the pytss code. For rapid coding we (and the user) can still use it, though. Removing the imports from `__init__.py` just means that you have to import the names via dot notation:
> 
> ```python
> from tpm2_pytss.types import *
> from tpm2_pytss.ESAPI import *
> # ...
> ```
> 
Yes, but that easily grows and requires an understanding of where everything is located, one example from this module is the Wrap function, it's not obvious where it's located without looking at the code. An external example is the cryptography module which would be hard to use without the extensive documentation.
 
> I understand that is can be slightly more effort to use the dot notation here, but I feel it adds clarity and it seems to be the python way to do things. There are probably good reasons but, to be frank, I am not the most experienced python expert either ;)
I agree on the point of clarity, but they aren't exclusive to each other and from an outside perspective it might not be easy to understand what to import.

I don't want to keep the rest of this PR from being merged so my suggestion would be so revisit the removal of ```__init__.py``` later on, after we have improved the documentation, but before the first release candidate for the new bindings.
By being strict about full path imports in the documentation we can encourage that usage as well.

williamcroberts:
> Yes, I threw out all asterisk imports in the pytss code. For rapid coding we (and the user) can still use it, though. Removing the imports from `__init__.py` just means that you have to import the names via dot notation:
> 
> ```python
> from tpm2_pytss.types import *
> from tpm2_pytss.ESAPI import *
> # ...
> ```
> 
> I can see that the imports in our own code are long-ish now (on my side, my pycharm hides them for me). I could easily change the formatting so that we do not have one import per line.
> 
> I understand that is can be slightly more effort to use the dot notation here, but I feel it adds clarity and it seems to be the python way to do things. There are probably good reasons but, to be frank, I am not the most experienced python expert either ;)

IIUC, it means that to use ESAPI I need to import ESAPI and types, which I feel is awkward. I just want to import the API and use it.

setup.cfg
```diff
@@ -43,6 +43,7 @@ dev =
     twine
     setuptools_scm[toml]>=3.4.3
     black==19.10b0
+    isort==5.9.1
```
williamcroberts, line 46: any reason we're pinning isort?

joholl:
> Yes, but that easily grows and requires an understanding of where everything is located, one example from this module is the Wrap function, it's not obvious where it's located without looking at the code. An external example is the cryptography module which would be hard to use without the extensive documentation.
 
Agreed. In the end, we need documentation for the API, anyway.

> I don't want to keep the rest of this PR from being merged so my suggestion would be so revisit the removal of __init__.py later on, after we have improved the documentation, but before the first release candidate for the new bindings.

I also agree. We probably don't want to advertise the wildcard imports in the documentation, but that's a conversation for another day.

I re-added the imports back to the `__init__.py` file. At the same time, I suggest using explicit imports (as opposed to wildcard imports) internally. This avoids naming conflicts (for instance `get_ptr()` is defined in `ESAPI` and `FAPI`).

test/test_esapi.py
```diff
@@ -3,9 +3,81 @@
...
+    TPMT_SYM_DEF,
+    TPMT_SYM_DEF_OBJECT,
+    TPMT_TK_AUTH,
+    TPMT_TK_CREATION,
+    TPMT_TK_HASHCHECK,
+    TPMT_TK_VERIFIED,
+    TPMU_HA,
+    TPMU_SYM_KEY_BITS,
+    TPMU_SYM_MODE,
+)
```
williamcroberts, line 79: Is this really better than `import *`. Why not just use the qualified names?
```python3
from tpm2_pytss import TCTI as TCTI, TSS2_Exception as TSS2_Exception, types
```

.ci/run.sh
```diff
@@ -66,10 +66,6 @@ function run_whitespace() {
   fi
 }
 
-function run_style() {
```
williamcroberts, line 69: Is their any reason style doesn't do both things? We also end up spinning an extra VM instance on GitHub which is expensive and takes more time then just running all are checks in one thing. We can change the logic to not fail fast, so we get everything that's wrong over the fist set of errors from the first tool being executed.

joholl: Unlike other linters, `isort` runs really quickly, yes. Style checking is not a bottle neck either way, so it doesn't really matter performance-wise. I find it more clean and organized to have one job per linter, especially if we want to add more linters in the future. That way you see instantly which linter failed and can instantly see the relevant log (instead of searching for the relevant bits).

test/test_esapi.py
joholl: I find using qualified names unnecessary long, but I do not have a strong opinion about this. Having `from tpm2_pytss.types import TPMU_SYM_KEY_BITS` is the default setting e.g. in `black`, `isort` and pycharm. It's just the way it's commonly done because it is non-ambiguous and reasonably short, I guess.

If you want to do it like `import tpm2_pytss.types` and `types.TPMU_SYM_KEY_BITS` every time, instead, that would be fine by me.

.ci/run.sh
williamcroberts: Its the VM spin-up time, nothing to do with whats executed. Also, we can only launch so many VMs at a time (offhand I think its 4). I'd rather dedicate parallel VM's to running actual function checks over additional "linter" checks. At some point in time, we will run multiple Python version checks before 1.0.

Right now we will have 3 nodes running just for style checks and if/when we add the linter, there's another instance.

What you can do is make each check it's own job, but run it in one node as multiple jobs. I think you can just mark the jobs as continue-on-error: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

setup.cfg
joholl: I just pinned it because black was pinned, as well. Besides that there is no particular reason. Do you want me to unpin it?

test/test_esapi.py
williamcroberts: > I find using qualified names unnecessary long, but I do not have a strong opinion about this. Having `from tpm2_pytss.types import TPMU_SYM_KEY_BITS` is the default setting e.g. in `black`, `isort` and pycharm. It's just the way it's commonly done because it is non-ambiguous and reasonably short, I guess.

Those are fine, and since they don't have a zillion classes, aren't overly terrible to deal with.

> 
> If you want to do it like `import tpm2_pytss.types` and `types.TPMU_SYM_KEY_BITS` every time, instead, that would be fine by me.

I'd rather that than an import list that's the majority of the file.

setup.cfg
williamcroberts: black is pinned for some bug that makes the formatting different and will cause issues if it's not that version. However, could probably be changed as well at some point to something `>=` once we identify the version thats borked.

williamcroberts: black is also only in beta-4, so it's probably best to pin it until its stable.

.ci/run.sh
joholl: Oh, that's a very good point! `continue-on-error` shows the stage as passed even though it failed, but I found another solution which continues on fail using `if: always()`.

test/test_esapi.py
joholl: Ok, as it turns out I made a promise too fast: [black cannot format multiple imports per line](https://github.com/psf/black/issues/204).

To sum this up: either it's a long import list in the tests or it is a wildcard import. I prefer the former because it is unambiguous and best practice.

If you want, I can re-add the wildcard imports (but keep isort for import formatting, i.e. alphabetic sorting etc.).

williamcroberts: > Ok, as it turns out I made a promise too fast: [black cannot format multiple imports per line](https://github.com/psf/black/issues/204).
> 
> To sum this up: either it's a long import list in the tests or it is a wildcard import. I prefer the former because it is unambiguous and best practice.

Yeah but sometimes it just doesn't make sense in certain situations. I think this is one of those situations. Perhaps a rule or guideline here where if the import list is over 10 things long or something like that....

Generally the reason you avoid wildcard imports are things like [this](https://rules.sonarsource.com/python/RSPEC-2208):
```
It could lead to conflicts between names defined locally and the ones imported.
It reduces code readability as developers will have a hard time knowing where names come from.
It clutters the local namespace, which makes debugging more difficult.
```

For item 1, the TPM isn't really usable without the types, you need em all.
For item 2, a 100 line import I think messes with readability more, and with IDE's the resolution of the class name is easily understood.
For item 3, yep it does, but see 1.

> 
> However, if you want, I can re-add the wildcard imports (but keep isort for import formatting, i.e. alphabetic sorting etc.).

Please do that. We can revisit explicit import lists later.

test/test_esapi.py
```diff
@@ -3804,7 +3876,3 @@ def test_PolicyTicket(self):
             self.ectx.PolicyTicket(
                 session, timeout, b"", b"", name, policy_ticket, session3=object()
             )
-
-
-if __name__ == "__main__":
-    unittest.main()
```
williamcroberts, line 3810: I like the ability to run these as standalone scripts.

test/test_types.py
```diff
@@ -6,7 +6,41 @@
 import itertools
 import unittest
 
-from tpm2_pytss import *
```
williamcroberts, line 9: keep this one as import * as well. The shorter lists can survive.

test/test_types.py
```diff
@@ -1297,7 +1331,3 @@ def test_TPML_DIGEST(self):
 
         with self.assertRaises(TypeError):
             TPML_PCR_SELECTION(TPML_AC_CAPABILITIES())
-
-
-if __name__ == "__main__":
-    unittest.main()
```
williamcroberts, line 1303: keep this

tpm2_pytss/ESAPI.py
```diff
@@ -2,12 +2,66 @@
 SPDX-License-Identifier: BSD-2
 """
 
-from ._libtpm2_pytss import lib
-
-from .types import *
```
williamcroberts, line 7: keep this wildcard one as well

tpm2_pytss/tsskey.py
```diff
@@ -3,11 +3,33 @@
 """
 
 import warnings
-from ._libtpm2_pytss import lib
-from .types import *
```
williamcroberts, line 7: keep this one...

setup.cfg
williamcroberts: unpin isort

williamcroberts (changes requested):

joholl:
Thanks for the review! I added the requested changes and rebased:

* Removed the big wildcard imports
* Unpinned isort
* Re-added `unittest.main()` at the end of tests (and added it to every test script)

williamcroberts:
@joholl if you want to get this into 1.0, can you rebase this. We're in the process of planning the release.

williamcroberts:
closing

PR #183, opened 2021-06-26
Esapi testing (PolicySecret and PolicyTicket)

williamcroberts:
Also, make getitem of a TPML_* of TPM2B_SIMPLE objects return the bytes themselves over the wrapper.

### DIFF
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -3647,6 +3647,66 @@ class TestEsys(TSS2_EsapiTest):
 
         self.assertEqual(tcti, self.ectx.tcti)
 
+    def test_PolicySecret(self):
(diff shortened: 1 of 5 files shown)


williamcroberts:
@whooo theirs a patch in here that when a getitem call occurs on a `TPML_<type>`, and the return is a TPM2B_SIMPLE_OBJECT, that it returns the bytes, thoughts on that?

whooo:
> @whooo theirs a patch in here that when a getitem call occurs on a `TPML_<type>`, and the return is a TPM2B_SIMPLE_OBJECT, that it returns the bytes, thoughts on that?

Sounds good to me!

PR #186, opened 2021-06-29
types: add equal comparison method for TPM2B_SIMPLE_OBJECT

whooo:
Simply pass the comparison to an instance of bytes

### DIFF
--- a/test/test_types.py
+++ b/test/test_types.py
@@ -1152,6 +1152,12 @@ class TypesTest(unittest.TestCase):
         b = bytes(dig)
         self.assertEqual(b, bob)
 
+        self.assertEqual(dig, bob)
+        self.assertNotEqual(dig, "pinchofbytes")
+
+        sdig = TPM2B_DIGEST(bob)
+        self.assertEqual(dig, sdig)
+
         with self.assertRaises(AttributeError):
             dig.size = 1
         with self.assertRaises(TypeError):
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -1100,6 +1100,10 @@ class TPM2B_SIMPLE_OBJECT(TPM_OBJECT):
         b = self.__bytes__()
         return binascii.hexlify(b).decode()
 
+    def __eq__(self, value):
+        b = self.__bytes__()
+        return b == value
+
 
 class TPML_Iterator(object):
     def __init__(self, tpml):


tpm2_pytss/types.py
```diff
@@ -1100,6 +1100,10 @@ def __str__(self):
         b = self.__bytes__()
         return binascii.hexlify(b).decode()
 
+    def __eq__(self, value):
+        b = self.__bytes__()
```
williamcroberts, line 1104: what if value is another instance of a TPM2B_SIMPLE_OBJECT? Can we just bytes(value) unconditionally to pick up comparisons between those and other "bytes like" objects?

whooo: I just checked it seems that ```__eq__``` for bytes does that for us:
```Python
>>> from tpm2_pytss.types import *
>>> d = TPM2B_DIGEST("kaka")
>>> s = TPM2B_DIGEST("baka")
>>> s == d
False
>>> s = TPM2B_DIGEST("kaka")
>>> s == d
True
>>> 
``` 

But I'll add a check for it in the test as well

test/test_types.py
```diff
@@ -1152,6 +1152,12 @@ def test_TPM2B_SIMPLE_OBJECT(self):
         b = bytes(dig)
         self.assertEqual(b, bob)
 
+        self.assertEqual(dig, bob)
+        self.assertNotEqual(dig, "pinchofbytes")
+
+        sdig = TPM2B_DIGEST(bob)
+        self.assertEqual(dig, sdig)
```
williamcroberts, line 1159: what if you pass `TPM2B_ATTEST(bob)`, we want that to fail.

whooo: why should it fail if it has the same value?

williamcroberts: My thought is that the type conveys information about the value, so the type should match. The API/Spec details different simple TPM2B's for context around the value, and that if the comparison involved two simple TPM2B's the instances should match around that context.

whooo: My understanding/view is that the ```__eq__``` method is just for value comparison (so ==) while "is" is used for checking the type as well.
So for example:
```Python
>>> b = b"1"
>>> ba = bytearray(b"1")
>>> b == ba
True
>>> b is ba
False
>>> b2 = b"1"
>>> b is b2
True
>>> 
```

williamcroberts: > My understanding/view is that the `__eq__` method is just for value comparison (so ==) while "is" is used for checking the type as well.
> So for example:
> 
> ```python
> >>> b = b"1"
> >>> ba = bytearray(b"1")
> >>> b == ba
> True
> >>> b is ba
> False
> >>> b2 = b"1"
> >>> b is b2
> True
> >>> 
> ```

Ahh ok, I forgot that `is` exists.

PR #188, opened 2021-06-30
Esapi testing finishing

williamcroberts:
@whooo with this PR It appears the full API is covered. Next ill go through and PEP8 the API to what @joholl was mentioning on #163. After that ill break the test_esapi.py into functional tests covering the main command as organized in the command specification. IE, CreatePrimary Tests will be in test_esapi_hiearchy_commands.py and things like test_Policy<XXX> will be in test_esapi_ea_commands.py.


whooo:
Great to hear, I'll focus on PEP8ifying and adding documentation for Wrap, MakeCredential and the different extra methods for some types (toPEM, fromPEM etc).

PR #207, opened 2021-08-10
wrap: simplify input parameters

williamcroberts:
Default the symkey to None, which will cause a key to be generated if
symdef is not None or not TPM2_ALG.NULL.

Default the symdef to aes128cfb which is the spec default protection
algorithm.

This way calls omitting the symdef and symkey will generate a wrapping
key and use aes128cfb be default.

### DIFF
--- a/tpm2_pytss/makecred.py
+++ b/tpm2_pytss/makecred.py
@@ -112,18 +112,26 @@ def make_credential(public, credential, name):
     return (credblob, secret)
 
 
-def wrap(newparent, public, sensitive, symkey, symdef):
+def wrap(newparent, public, sensitive, symkey=None, symdef=None):
     """Wraps key under a TPM key hierarchy
 
     Args:
         newparent (TPMT_PUBLIC): The public area of the parent
         public (TPM2B_PUBLIC): The public area of the key
         sensitive (TPM2B_SENSITIVE): The sensitive area of the key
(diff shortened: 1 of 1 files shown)


whooo:
I would default symdef to None to make inner encryption/integrity off by default as outer encryption/integrity is always done by the wrap function

williamcroberts:
> I would default symdef to None to make inner encryption/integrity off by default as outer encryption/integrity is always done by the wrap function

I was leaning towards on by default over off by default but I can go the other way.

whooo:
> > I would default symdef to None to make inner encryption/integrity off by default as outer encryption/integrity is always done by the wrap function
> 
> I was leaning towards on by default over off by default but I can go the other way.

I'm OK with both, but I don't think inner encryption adds much in most cases, regardless it makes using wrap much easier.

williamcroberts:
> > > I would default symdef to None to make inner encryption/integrity off by default as outer encryption/integrity is always done by the wrap function
> > 
> > 
> > I was leaning towards on by default over off by default but I can go the other way.
> 
> I'm OK with both, but I don't think inner encryption adds much in most cases, regardless it makes using wrap much easier.

I'm going to head your way with it as I agree

williamcroberts:
@whoo this look better?

whooo:
> @whoo this look better?

Looks good!

PR #208, opened 2021-08-11
crypto: add support for getting the public part from private keys

whooo:
Both RSA and ECC private keys can provide the public key so this enables from_pem in TPMT_PUBLIC/TPM2B_PUBLIC on private keys.

### DIFF
--- a/tpm2_pytss/crypto.py
+++ b/tpm2_pytss/crypto.py
@@ -80,7 +80,7 @@ def _int_to_buffer(i, b):
     b.buffer = i.to_bytes(length=s, byteorder="big")
 
 
-def key_from_encoding(data):
+def key_from_encoding(data, password=None):
     sdata = data.strip()
     key = None
     if sdata.startswith(b"-----BEGIN CERTIFICATE-----"):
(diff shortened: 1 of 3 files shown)


tpm2_pytss/crypto.py
```diff
@@ -88,6 +88,11 @@ def key_from_encoding(data):
         key = cert.public_key()
     elif sdata.startswith(b"-----BEGIN PUBLIC KEY-----"):
         key = load_pem_public_key(data, backend=default_backend())
+    elif sdata.startswith(b"-----BEGIN RSA PRIVATE KEY-----") or sdata.startswith(
+        b"-----BEGIN EC PRIVATE KEY-----"
+    ):
+        pkey = load_pem_private_key(data, password=None, backend=default_backend())
```
williamcroberts, line 94: We probably want to plumb password through

tpm2_pytss/crypto.py
```diff
@@ -88,6 +88,11 @@ def key_from_encoding(data):
         key = cert.public_key()
     elif sdata.startswith(b"-----BEGIN PUBLIC KEY-----"):
         key = load_pem_public_key(data, backend=default_backend())
+    elif sdata.startswith(b"-----BEGIN RSA PRIVATE KEY-----") or sdata.startswith(
+        b"-----BEGIN EC PRIVATE KEY-----"
```
williamcroberts, line 92: Not a fix now type of thing, but IIRC, scissor lines in PEM are not required unless you have N>1 things in a single file. So it would probably be best to just try/except the whole thing like the DER code.

williamcroberts (changes requested):

tpm2_pytss/crypto.py
whooo: sure, I'll add a password argument to from_pem for both the public and private versions

tpm2_pytss/crypto.py
whooo: So more or less skip the sdata.startswith and just do try ... except pass?

williamcroberts: Yeah but you don't have to put it in this PR, ill take it without it.

tpm2_pytss/crypto.py
whooo: Done

williamcroberts (approved):

PR #211, opened 2021-08-13
ci: github action for package build/twine upload

### LINKED ISSUE #108: test twine with test pypi
Test the release process using https://packaging.python.org/guides/using-testpypi/

terriko:
* fixes #108  

This sets up a Github Actions job to build the package and test an upload to testpypi using twine.

* The test pyPI credential secret I'm using doesn't exist on the main repo, but I figured it was easier to open it for review this way.
* I've omitted the wheel uploads because the wheel created right now isn't a valid manylinux wheel. (you might want those later but one problem at a time)
* TestPyPI doesn't like localparts in versions, so I tweaked the config in pyproject.toml to disable the local version.  You might want to tweak this differently because the way I set it up means you'll get version overlap, and testPyPI doesn't accept duplicate versions.  I'm guessing how we fix this will depend on how we want this Github Action to be triggered
* The Action is currently being triggered on push, which likely isn't desirable, especially given the duplicate versions problem.  
* We may want to change the upload so that part is allowed to fail if we aren't creating unique versions.


.github/workflows/twine.yaml
```diff
@@ -0,0 +1,66 @@
...
+          sudo make install
+      - name: Get tpm2-pytss package
+        uses: actions/checkout@v2
+      - name: Build tpm2-pytss package
+        # Github Actions doesn't need
+        # PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
+        run: python -m build
+      - name: Publish to TestPyPi
+        env:
+          PUBLISH_PKG: 1
```
williamcroberts, line 61: Does this ENV var actually get used?

.github/workflows/twine.yaml
```diff
@@ -0,0 +1,66 @@
...
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.TESTPYPI_TOKEN }}
+        # Don't upload the .whl files, not manylinux compatible
+        run: python3 -m twine upload --repository testpypi dist/*.tar.gz
```
williamcroberts, line 65: Should we modify .ci/run.sh for the PUBLISH: 1 (PUBLISH=1) function?

PR #213, opened 2021-08-16
[RFC] types: add support for loading HMAC keys and seal data from bytes

whooo:
The reason for only having the keyedhash_from_secret method on TPMT_SENSITIVE / TPM2B_SENSITIVE is that the private and public parts are very interconnected.

I'm not 100% sure about this API, so some feedback would be appreciated.
Especially the seed value handling is a bit icky, I'm not sure that all zeros is a sane default or if it should be random.

After working out the API I'll do something similar for symmetric ciphers.

### DIFF
--- a/test/test_crypto.py
+++ b/test/test_crypto.py
@@ -665,3 +665,45 @@ class CryptoTest(TSS2_EsapiTest):
 
         with self.assertRaises(ValueError):
             TPMT_PUBLIC.from_pem(ecc_encrypted_key, password=b"passpass")
+
+    def test_keyedhash_from_secret(self):
+        secret = b"secret key"
+        scheme = TPMT_KEYEDHASH_SCHEME(scheme=TPM2_ALG.HMAC)
+        scheme.details.hmac.hashAlg = TPM2_ALG.SHA256
+        (sens, pub) = TPM2B_SENSITIVE.keyedhash_from_secret(secret, scheme=scheme)
+
+        self.assertEqual(pub.publicArea.type, TPM2_ALG.KEYEDHASH)
+        self.assertEqual(pub.publicArea.nameAlg, TPM2_ALG.SHA256)
+        self.assertEqual(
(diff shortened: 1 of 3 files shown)


williamcroberts:
> The reason for only having the keyedhash_from_secret method on TPMT_SENSITIVE / TPM2B_SENSITIVE is that the private and public parts are very interconnected.
> 
> I'm not 100% sure about this API, so some feedback would be appreciated.
> Especially the seed value handling is a bit icky, I'm not sure that all zeros is a sane default or if it should be random.
> 
> After working out the API I'll do something similar for symmetric ciphers.

TL;DR need to specify or random generated to prevent known data hash attacks (ie a rainbow table).

So I wasn't sure off hand myself, I was leaning towards randomly generated, but I wanted a better reason than my feelings. So, I did some digging today and this was in the spec:

```
There are two reasons for generating the unique parameter for symmetric objects in this way. The first is
that it protects the contents of the user-provided data. If the secret data has low entropy, then making the
unique parameter a simple digest of that data would allow an offline attack to determine what the secret
data might be. The large, random, obfuscation value generated by the TPM is not known to an attacker,
which mitigates this threat.
```

Which if you look at how unique is calculated for symmetric objects, makes perfect sense. Since it's just the hash of the sensitive data itself + seed. So the seed is the salt when storing something like user passwords.

Looking at the TPM code itself, if it's a TPM2_ALG_SYMCIPHER or TPM2_ALG_KEYEDHASH it always generates a seed, and this we should too.

williamcroberts:
Section 27.5.3.1 of https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part1_Architecture_pub.pdf

whooo:
Of course, that makes perfect sense, updated the code to generate some random bytes if seed is not passed to the method

williamcroberts (approved):

PR #216, opened 2021-09-03
TPM signature verification function

whooo:


tpm2_pytss/crypto.py
```diff
@@ -291,3 +292,75 @@ def get_digest_size(alg):
...
+def verify_signature_rsa(signature, key, data):
+    dt = _get_digest(signature.signature.rsapss.hash)
+    if dt is None:
+        raise ValueError(
+            f"unsupported digest algorithm: {signature.signature.rsapss.hash}"
+        )
+    if signature.sigAlg == lib.TPM2_ALG_RSASSA:
+        pad = padding.PKCS1v15()
+    elif signature.sigAlg == lib.TPM2_ALG_RSAPSS:
+        pad = padding.PSS(mgf=padding.MGF1(dt()), salt_length=dt.digest_size)
```
williamcroberts, line 306: TL;DR Check with `slen == hlen` if fail try `slen == PSS.MAX_LENGTH`

Oh PSS on TPMs.... So salt_len is often maximized and not equal to hash length. This has been a nightmare to deal with in tpm2-pkcs11 code. If the FIPS bit is set it's required to be the largest size allowed by FIPS which in most cases is slen == hlen. If it's not set slen == MAX. I newer specs and TPMs, hlen == slen, but older TPM's may continue the old behavior.

tpm2_pytss/crypto.py
```diff
@@ -291,3 +292,75 @@ def get_digest_size(alg):
...
+    else:
+        raise ValueError(f"unsupported RSA signature algorihtm: {signature.sigAlg}")
+
+    sig = bytes(signature.signature.rsapss.sig)
+    key.verify(sig, data, pad, dt())
+
+
+def verify_signature_ecc(signature, key, data):
+    dt = _get_digest(signature.signature.ecdsa.hash)
```
williamcroberts, line 315: `.any.hash` again

williamcroberts (changes requested):

tpm2_pytss/crypto.py
whooo: I see, I added an simple trial and error solution

williamcroberts (approved):

PR #218, opened 2021-09-05
Make FAPI ready for release

### LINKED ISSUE #54: Add FAPI

joholl:
Changes:

* rename api files and classes to uppercase (pep8), as discussed in #54
* add optional size param to `FAPI.create_seal()`
* test Fapi against both RSA and ECC
* account for yet another bug in fapi: tpm2-software/tpm2-tss/issues/2092

Fixes #54

The tests are still brittle because they test against the system FAPI configs (and the defaults are assumed). This should not be a showstopper for the first release, though. Also, the documentation can be improved, for sure. However, to get this shipped, I'd say FAPI is ready for release. 

Edit: Fyi, I'm gonna be on leave the following week.

### DIFF
--- a/test/test_fapi.py
+++ b/test/test_fapi.py
@@ -1035,6 +1035,10 @@ class TestFapiRSA(Common):
         assert tpm_2b_private.size == 0xDE
         assert policy == ""
 
+    @pytest.mark.skipif(
+        not is_bug_fixed(fixed_in="3.2", backports=["2.4.7", "3.0.5", "3.1.1"]),
+        reason="tpm2-tss bug, see #2092",
+    )
     def test_encrypt_decrypt(self, decrypt_key):
         plaintext = b"Hello World!"
         ciphertext = self.fapi.encrypt(decrypt_key, plaintext)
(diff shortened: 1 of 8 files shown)


test/test_fapi.py
```diff
@@ -530,11 +526,21 @@ def test_nv_write_read(self, nv_ordinary):
         assert log == ""
 
     def test_nv_increment(self, nv_increment):
+        # TODO initial increment should not be necessary, check in with upstream
```
williamcroberts, line 529: Did you file a bug?

joholl: No, I think this is not a bug, but another TPM spec quirk. [III. Command Spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf), 31.13 TPM2_NV_Read:

> If the NV Index has been  defined but the `TPMA_NV_WRITTEN` attribute is `CLEAR`, then this command 
shall return `TPM_RC_NV_UNINITIALIZED` [...].

That is, if the index was defined but never written to (here incremented), reading fails. That's counter-intuitive, so we might want to add a more verbose warning.

williamcroberts (approved):

PR #220, opened 2021-09-29
Add mssing typing

williamcroberts:
This adds some cleanups and missing typing. I did not do crypto.py because it imparts a circular dependency on types.py. Not sure what the best way to break that is.

[crypto.patch.txt](https://github.com/tpm2-software/tpm2-pytss/files/7254756/crypto.patch.txt)


whooo:
> This adds some cleanups and missing typing. I did not do crypto.py because it imparts a circular dependency on types.py. Not sure what the best way to break that is.
> 
> [crypto.patch.txt](https://github.com/tpm2-software/tpm2-pytss/files/7254756/crypto.patch.txt)

using strings for the types might work, so something like obj: 'types..TPMT_PUBLIC'

williamcroberts:
> > This adds some cleanups and missing typing. I did not do crypto.py because it imparts a circular dependency on types.py. Not sure what the best way to break that is.
> > [crypto.patch.txt](https://github.com/tpm2-software/tpm2-pytss/files/7254756/crypto.patch.txt)
> 
> using strings for the types might work, so something like obj: 'types..TPMT_PUBLIC'

Not following, can you show me in a little patch (just do one)?

whooo:
> Not following, can you show me in a little patch (just do one)?

Something like:

```
diff --git a/tpm2_pytss/crypto.py b/tpm2_pytss/crypto.py
index 18409d7..eb13aaf 100644
--- a/tpm2_pytss/crypto.py
+++ b/tpm2_pytss/crypto.py
@@ -182,7 +182,7 @@ def private_from_encoding(data, obj, password=None):
         raise RuntimeError(f"unsupported key type: {key.__class__.__name__}")
 
 
-def public_to_key(obj):
+def public_to_key(obj: 'types.TPMT_PUBLIC'):
     key = None
     if obj.type == lib.TPM2_ALG_RSA:
         b = obj.unique.rsa.buffer

```

williamcroberts:
> > Not following, can you show me in a little patch (just do one)?
> 
> Something like:
> 
> ```
> diff --git a/tpm2_pytss/crypto.py b/tpm2_pytss/crypto.py
> index 18409d7..eb13aaf 100644
> --- a/tpm2_pytss/crypto.py
> +++ b/tpm2_pytss/crypto.py
> @@ -182,7 +182,7 @@ def private_from_encoding(data, obj, password=None):
>          raise RuntimeError(f"unsupported key type: {key.__class__.__name__}")
>  
>  
> -def public_to_key(obj):
> +def public_to_key(obj: 'types.TPMT_PUBLIC'):
>      key = None
>      if obj.type == lib.TPM2_ALG_RSA:
>          b = obj.unique.rsa.buffer
> ```
> 
> Reference: https://www.python.org/dev/peps/pep-0484/#forward-references

That works... I literally spent 20 mins looking for how to do forward references in Python, thank you.

PR #227, opened 2021-10-11
Support tss under 3

williamcroberts:
Support tpm2-tss 2.4.0+


williamcroberts:
So it seems that install tpm2-tss version 2.4.0 is breaking something that I can't quite replicate on my end.

PR #228, opened 2021-10-12
ESAPI: None to mean empty TPM2B_SESNTIVE_CREATE

williamcroberts:
in TPM2 object creation commands (create_primary, create and
create_loaded), support None for TPM2B_SENSITIVE_CREATE so creating
simple passwordless objects is dead simple.


whooo:
how about making in_sensitive optional as well?

Issue #235, opened 2021-10-13
support tools YAML output


williamcroberts:
For things like tpm2b_public, support the YAML structured output of tools. Find ALL YAML outputs and have data type conversions for them via a to_yaml call.

whooo:
I might take this on, it correlates to some other things I'm looking into

whooo:
Closing as support was merged via https://github.com/tpm2-software/tpm2-pytss/pull/327

PR #240, opened 2021-10-14
Serialization

whooo:
Adding functionality to be able to serialize different TPM structures

### DIFF
--- a/test/test_encoding.py
+++ b/test/test_encoding.py
@@ -1,0 +1,694 @@
+#!/usr/bin/python3 -u
+"""
+SPDX-License-Identifier: BSD-2
+"""
+import unittest
+
+from tpm2_pytss import *
+from tpm2_pytss.encoding import (
+    base_encdec,
+    json_encdec,
+)
+
+
+class SerializationTest(unittest.TestCase):
+    def test_base_simple_tpm2b(self):
+        enc = base_encdec()
+        dig = TPM2B_DIGEST(b"falafel")
+        ev = enc.encode(dig)
+        self.assertEqual(ev, "66616c6166656c")
+
+    def test_base_friendly_int(self):
+        enc = base_encdec()
+        ev = enc.encode(TPM2_ALG.SHA256)
+        self.assertEqual(ev, 0x0B)
+
+    def test_base_friendly_intlist(self):
+        enc = base_encdec()
+        ev = enc.encode(TPMA_OBJECT(TPMA_OBJECT.RESTRICTED | TPMA_OBJECT.FIXEDTPM))
+        self.assertEqual(ev, 0x10002)
+
+    def test_base_int(self):
+        enc = base_encdec()
+        ev = enc.encode(1337)
+        self.assertEqual(ev, 1337)
+
+    def test_base_complex_tpm2b(self):
+        enc = base_encdec()
+        p = TPM2B_ECC_POINT(TPMS_ECC_POINT(x=b"\x01", y="\x02"))
+        ev = enc.encode(p)
+        self.assertEqual(ev, {"x": "01", "y": "02"})
+
+    def test_base_tpml(self):
+        enc = base_encdec()
+        al = TPML_ALG((TPM2_ALG.SHA256, TPM2_ALG.SHA384, TPM2_ALG.SHA512))
+        ev = enc.encode(al)
+        self.assertEqual(ev, [TPM2_ALG.SHA256, TPM2_ALG.SHA384, TPM2_ALG.SHA512])
+
+    def test_base_TPMS_CAPABILITY_DATA(self):
+        enc = base_encdec()
+        algs = TPMS_CAPABILITY_DATA(capability=TPM2_CAP.ALGS)
+        algs.data.algorithms = TPML_ALG_PROPERTY(
+            (TPMS_ALG_PROPERTY(alg=TPM2_ALG.SHA256, algProperties=TPMA_ALGORITHM.HASH),)
+        )
+        ev = enc.encode(algs)
+        self.assertEqual(
+            ev, {"capability": 0, "data": [{"alg": 11, "algProperties": 4}]}
+        )
+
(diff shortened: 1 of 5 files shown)


whooo:
While this doesn't exactly cover https://github.com/tpm2-software/tpm2-pytss/issues/235 it might be useful as a base.

tpm2_pytss/constants.py
```diff
@@ -1267,6 +1310,16 @@ class TPMA_CC(TPM_FRIENDLY_INT):
     RES_MASK = lib.TPMA_CC_RES_MASK
     RES_SHIFT = lib.TPMA_CC_RES_SHIFT
 
+    @property
+    def commandindex(self) -> int:
+        """int: The command index"""
+        return self & self.COMMANDINDEX_MASK
+
+    @property
+    def chandles(self) -> int:
```
williamcroberts, line 1319: do we want:

```python
def is_flushed()
def is_vendor()
def is_nv()
def is_extensive()
def has_rhandle()
... etc ...
```

whooo: Possibly, could be useful with the different bits of TPMA_NV and TPMA_OBJECT for example

test/test_encoding.py
```diff
@@ -0,0 +1,743 @@
...
+    base_encdec,
+    toolslike_encdec,
+    json_encdec,
+)
+
+
+class SerializationTest(unittest.TestCase):
+    def test_base_simple_tpm2b(self):
+        enc = base_encdec()
+        dig = TPM2B_DIGEST(b"falafel")
```
williamcroberts, line 18: im hungry now

test/test_encoding.py
```diff
@@ -0,0 +1,743 @@
+#!/usr/bin/python3 -u
+"""
+SPDX-License-Identifier: BSD-2
+"""
+import unittest
+
+from tpm2_pytss import *
+from tpm2_pytss.encoding import (
+    base_encdec,
```
williamcroberts, line 9: So what's base_encdec all hex strings? If so perhaps call this hex_encdec

test/test_encoding.py
```diff
@@ -0,0 +1,743 @@
...
+    toolslike_encdec,
```
williamcroberts, line 10: Is it supposed to be tools like or actually match the tools? If it should match the tools, lets not be ambiguous about that and call it tpm2tools_encdec.

Plus pep8 classes should be camelCased.

test/test_encoding.py
```diff
@@ -0,0 +1,743 @@
...
+    json_encdec,
```
williamcroberts, line 11: I'm assuming this matches the json spec?
  - https://trustedcomputinggroup.org/resource/tcg-tss-json/

test/test_encoding.py
whooo: it's the base class for other encoders/decoders, so it's the minimal viable implementation so subclasses can just implement the needed changes

test/test_encoding.py
whooo: That's the idea

test/test_encoding.py
williamcroberts: Whats the difference, ordering or missing fields?

test/test_encoding.py
williamcroberts: Is it supposed to be public? If it's public the name should be more descriptive of what it does IMO.

test/test_encoding.py
whooo: different field names and tpm2-tools flatten some structures, for example tpm2_readpublic puts the RSA parameters at the same level as type and nameAlg, so toolslike_encdec should be seen more as TPMA_FRIENDLY_INTLIST.parse then a compatibility layer for the tools

williamcroberts: > different field names and tpm2-tools flatten some structures, for example tpm2_readpublic puts the RSA parameters at the same level as type and nameAlg, so toolslike_encdec should be seen more as TPMA_FRIENDLY_INTLIST.parse then a compatibility layer for the tools

Folks over in tpm.dev are clamoring for a way to get tools YAML in and out to match. Is their anyway we can add in the special handling for tools?

test/test_encoding.py
whooo: I would prefer it to be public, if users want to implement their own encoders/decoders
I could always rename it to minimal_encdec and improve the description in the docstring

test/test_encoding.py
whooo: It's possible, I wouldn't do that in this class tho
It's just a lot of work going through all the source code for the tools and implementing + verifying it

williamcroberts: > It's possible, I wouldn't do that in this class tho It's just a lot of work going through all the source code for the tools and implementing + verifying it

Let's take base and json now, and then start a longer running branch for tools.

williamcroberts: FYI the tools YAML support is indicated in ticket: https://github.com/tpm2-software/tpm2-pytss/issues/235

whooo: > 
> Let's take base and json now, and then start a longer running branch for tools.

I would like to keep toolslike as well, as I have at least two use cases for that
I'll see if I can create a new branch of this and start the tools-part there in a couple of days

williamcroberts: > > Let's take base and json now, and then start a longer running branch for tools.
> 
> I would like to keep toolslike as well, as I have at least two use cases for that I'll see if I can create a new branch of this and start the tools-part there in a couple of days

The problem with tools-like, is that if we merge it and don't make it tools-firm by release, we have effectively made a new format. We already have enough formats to deal with. If you want to hide it behind a feature flag where to use it you have to buy into the fact it's experimental, that's fine too. You could just make it _toolslike_encdec indicating that it's private still and you probably shouldn't use it, unless you know what your doing, like you do.

whooo: I see your point, I'll remove it

williamcroberts (approved):

Issue #248, opened 2021-10-20
restructuring


williamcroberts:
Thoughts on restructuring the code, current issues:
- Cyclic dependencies, for instance types and crypto. Crypto had to be dsigned to be type agnostic.
- Some things that should be "private" are not marked with an "_" prefix

Things to consider:
A package or module prefixed with an _ so we don't have to mark every function

whooo:
How about a subdirectory called (_)internal for the private code?
And then:
* Move generate_seed, hmac and encrypt to crypto
* Move crypto to internals
* Move make_credential and wrap to a module called utils.py (which can container other smaller function/classes such as EK generation)
* Move all TPM_FRIENDLY_INT code to constants.py

Issue #242, opened 2021-10-14
Should crypto methods like wrap go into crypto.py and not makecred.py


williamcroberts:
So their are various crypto routines in makecred.py:
  - wrap
  - encrypt
  - hmac
  - generate_seed_<xxx>

Should these go to crypto.py?

With just make_cred left, should that just go to crypto as well and we can drop makecred.py?

@whooo  thoughs?

whooo:
I think this can be closed now as it was part of the restructure

PR #251, opened 2021-10-26
restructure

whooo:
Regarding https://github.com/tpm2-software/tpm2-pytss/issues/248

- [X] Move friendly ints to constants.py
- [X] Move generate_seed, hmac and encrypt from makecred to crypto
- [X] Move crypto to internal
- [X] move current utils to internal
- [X] Move make_credential and wrap from makecred to utils
- [X] Use constants in crypto instead of lib

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/248 and https://github.com/tpm2-software/tpm2-pytss/issues/242


whooo:
@joholl, could you have a look at https://github.com/tpm2-software/tpm2-pytss/pull/251/commits/62756dee4283fec91bff0fc8365c38df1771ef43 and see if it makes sense? the moved code does mostly seem to be used internally but I might have missed something.

tpm2_pytss/internal/crypto.py
```diff
@@ -345,34 +348,100 @@ def verify_signature_hmac(signature, key, data):
...
+    if halg is None:
+        raise ValueError(f"unsupported digest algorithm {hashAlg}")
+    seed = secrets.token_bytes(halg.digest_size)
+    mgf = padding.MGF1(halg())
+    padd = padding.OAEP(mgf, halg(), label)
+    enc_seed = key.encrypt(seed, padd)
+    return (seed, enc_seed)
+
+
+def generate_ecc_seed(
```
williamcroberts, line 397: Why are the 2 seed functions non-underscore versus underscore?

Is your convention here that things under internal are all for internal use, but functions/methods that start with an `_` are not general internal API. Ie something like underscore prefix functions are static functions for use in file scope and non-underscore prefix internal API is for general internal API consumption?

williamcroberts (approved):
Generally LGTM, just a question. Can you rebase and update?

tpm2_pytss/internal/crypto.py
whooo: I skipped them as they aren't imported anywhere, they are only called by _generate_seed.
I'll fix it when I rebase

Issue #255, opened 2021-10-27
tpm2-pytss crashes with FAPI.quote


gaetanww:
tpm2-pytss is trying to create a string from a null pointer when returning from Fapi_Quote. See below:

```
<ipython-input-6-56c42e72ab86> in <module>
----> 1 out = fapi.quote('HE/KEY', [0, 1, 2, 3, 4, 5, 6, 7, 8, 9 , 10])

/opt/zte_py/.venv/lib/python3.8/site-packages/tpm2_pytss/FAPI.py in quote(self, path, pcrs, quote_type, qualifying_data)
    953             bytes(ffi.unpack(get_dptr(signature, lib.Fapi_Free), signature_len[0])),
    954             ffi.string(get_dptr(pcr_log, lib.Fapi_Free)).decode(),
--> 955             ffi.string(get_dptr(certificate, lib.Fapi_Free)).decode(),
    956         )
    957 
RuntimeError: cannot use string() on <cdata 'char *' NULL>
```

To reproduce:
```python
from tpm2_pytss import *
tcti = 'swtpm'
FAPIConfig(profile_name='P_ECCP256SHA256', tcti=tcti, ek_cert_less='yes')
fapi = FAPI()
fapi.provision()
fapi.create_key('HE/KEY', 'system, sign, restricted')
out = fapi.quote('HE/KEY', [0, 1, 2, 3, 4, 5, 6, 7, 8, 9 , 10])
fapi.close()
```

whooo:
Can you test with the fix in https://github.com/tpm2-software/tpm2-pytss/pull/258 ?

williamcroberts:
@gaetanww any update, we merged the fixes would like confirmation before closing the ticket.

gaetanww:
@williamcroberts I tested and discussed the fix in #258. No crash for me anymore!

Issue #252, opened 2021-10-26
Wrong return type `create_primary`


gaetanww:
`ESAPI.create_primary` has the wrong return type:
[type hint](https://github.com/tpm2-software/tpm2-pytss/blob/master/tpm2_pytss/ESAPI.py#L4798) does not match [actual return value](https://github.com/tpm2-software/tpm2-pytss/blob/master/tpm2_pytss/ESAPI.py#L4880).

This kind of error could be spotted by mypy, is it worth running it in CI?

williamcroberts:
Fixed by PR #258

PR #258, opened 2021-11-02
Small bug fixes

whooo:
Fixes for
* https://github.com/tpm2-software/tpm2-pytss/issues/255
* https://github.com/tpm2-software/tpm2-pytss/issues/252

### DIFF
--- a/tpm2_pytss/FAPI.py
+++ b/tpm2_pytss/FAPI.py
@@ -957,7 +957,9 @@ class FAPI:
             ffi.string(get_dptr(quote_info, lib.Fapi_Free)).decode(),
             bytes(ffi.unpack(get_dptr(signature, lib.Fapi_Free), signature_len[0])),
             ffi.string(get_dptr(pcr_log, lib.Fapi_Free)).decode(),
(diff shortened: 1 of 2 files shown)


tpm2_pytss/FAPI.py
```diff
@@ -957,7 +957,9 @@ def quote(
             ffi.string(get_dptr(quote_info, lib.Fapi_Free)).decode(),
             bytes(ffi.unpack(get_dptr(signature, lib.Fapi_Free), signature_len[0])),
             ffi.string(get_dptr(pcr_log, lib.Fapi_Free)).decode(),
```
gaetanww, line 951: Is it worth checking for NULL pointer here too? tss2-fapi returns a '\n' for now but it could change in the future and crash tpm2-pytss again.

whooo: For the future I guess we need to check all FAPI calls that might return NULL

williamcroberts: Maybe make get_dptr smarter?

Issue #241, opened 2021-10-14
objects with friendly int fields should return the correct type


whooo:
For example, objectAttributes and type in a TPMT_PUBLIC should be a TPM2_ALG and a TPMA_OBJECT instance, not an ints

williamcroberts:
Anything should return the right type

williamcroberts:
@whooo isn't this fixed by PR #246 and we can close this?

whooo:
> @whooo isn't this fixed by PR #246 and we can close this?

Not really, it's also about the right type for object fields, for example, getting sigAlg from a TPMT_SIGNATURE instance should return a TPM2_ALG, not an ordinary int.
As discussed in https://github.com/tpm2-software/tpm2-pytss/pull/240

williamcroberts:
Ahh yes, I recall now. This was generating some type of type info so we can cast the int's back into their wrapper classes.

PR #262, opened 2021-11-15
add code to generate mapping between integer fields TPM2_FRIENDLY_INT…

whooo:
… types

This will parse tss2_tpm2_types.h during the build process and change
parts of tpm2_pytss/internal/type_mapping.py

Empty _type_map and _element_type_map are included so running unit
tests will continue to work.

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/241


whooo:
It's a draft PR as I haven't been able to get ```pip install .``` to work yet but it would be good with some feedback regardless

whooo:
Might be worth adding it to the sdist command as well

setup.py
```diff
@@ -1,8 +1,163 @@
...
+from pkgconfig import pkgconfig
+from pycparser import c_parser, preprocess_file
+from pycparser.c_ast import Decl, Typedef, TypeDecl, IdentifierType, Struct, ArrayDecl
 
 # workaround bug https://github.com/pypa/pip/issues/7953
 site.ENABLE_USER_SITE = "--user" in sys.argv[1:]
 
-setup(use_scm_version=True, cffi_modules=["scripts/libtss2_build.py:ffibuilder"])
+
+class type_generator(build_py):
```
williamcroberts, line 14: Is their a way we can put this class somewhere else and import it into setup.py?

setup.py
```diff
@@ -1,8 +1,163 @@
...
+        "TPMI_RH_OWNER": "TPM2_RH",
+        "TPMI_RH_ENDORSEMENT": "TPM2_RH",
+        "TPMI_RH_PROVISION": "TPM2_RH",
+        "TPMI_RH_CLEAR": "TPM2_RH",
+        "TPMI_RH_NV_AUTH": "TPM2_RH",
+        "TPMI_RH_LOCKOUT": "TPM2_RH",
+        "TPMI_RH_NV_INDEX": "TPM2_RH",
+        "TPMI_RH_AC": "TPM2_RH",
+        "TPMI_RH_ACT": "TPM2_RH",
+    }
```
williamcroberts, line 50: Why the cares and the type_mapping dictionaries? What information do we need that we're missing?

setup.py
```diff
@@ -1,8 +1,163 @@
...
+            for l in tf.readlines():
+                if l == "_type_map = {}\n":
+                    mout = mout + mstr
+                elif l == "_element_type_map = {}\n":
+                    mout = mout + estr
+                else:
+                    mout = mout + l
+            tf.seek(0)
+            tf.truncate(0)
+            tf.write(mout)
```
williamcroberts, line 156: Why do we need to read in the file and then update it? Don't we just take the generated map and dump it to the file no matter what? This would be a file we don't include in the source tree and add to .gitignore.

setup.py
whooo: I tried that but that was the reason that pip install failed, didn't find a solution for it

williamcroberts: > I tried that but that was the reason that pip install failed, didn't find a solution for it

That's a good enough reason for me :-p. That's a refinement that can always happen later and isn't a blocker.

setup.py
whooo: cares are the typedefs that it checks for (including child typedefs), type_mapping is overriding some, for example TPMI_RH_OWNER is a typedef of TPM2_HANDLE, but I found TPM2_RH to be a better match

setup.py
whooo: The primary reason I choose that is so we don't need to catch any import exceptions in types.py, but I'm fine with either way

whooo:
I should note that I'm not too happy with this as the build and the source will differ to much, for example the unit tests will behave different depending on if it's run against a build or just the source.
While I'm OK with parsing the C code perhaps this should be done as part of a release process? So use the same code to generate type_mapping.py but not during every build?

whooo:
> > I should note that I'm not too happy with this as the build and the source will differ to much, for example the unit tests will behave different depending on if it's run against a build or just the source.
> 
> I don't think its possible to run the unit tests against a non-built source, as CFFI won't be setup properly. I'm assuming you mean something like a repo that has a git clean -fdx run and run something like pytest?
> 
The generated code is stored in the build directory, but I found a workaround to generate it during the build in the source directory. Just need to fix all the affected tests and make sure the file isn't included in the source distribution.

> I'm failing to understand the apprehension here? I see this as a common paradigm that gets reused to deal with monotonous code that is tough to maintain when it needs to stay in lockstep with some other set of code.
> 
I should have been more clear, I don't mind generating the code, it's about this implementation,

> Another interesting future feature would be using it to generate all the type classes and constants so all we have to implement is the extras and the base classes.
>
And embedding the version of tpm2-tss so pkgconfig isn't required during runtime

setup.py
whooo: I changed it so it will overwrites the whole file

whooo:
sorry for the spam, but the parser doesn't seem to work in the CI flow for an unknown reason

whooo:
@williamcroberts, I'm happy enough with this now.
Regarding the LGTM alert, I rather live with the alert, the other two solutions are to import everything from constants using a wildcard or prefix every TPMA_OBJECT, TPM2_ALG and TPM2_ECC_CURVE reference with constants

williamcroberts:
> @williamcroberts, I'm happy enough with this now. Regarding the LGTM alert, I rather live with the alert, the other two solutions are to import everything from constants using a wildcard or prefix every TPMA_OBJECT, TPM2_ALG and TPM2_ECC_CURVE reference with constants

If we're generating just prefix it with `constants` or import it `as` so you can have a short form. But since it's generated, i'd just long form it IMO. But i'm open to why we don't want to do that.

setup.py
williamcroberts: We want it to match what the original typedef is from the C source, no?
`tss2_tpm2_types.h:850:typedef TPM2_HANDLE TPMI_RH_OWNER;`

Also, if we're looking at a struct to get it's original type, isn't that "cares". I don't still get why we need to maintain a static map. In my head the logic is:
1. look at struct/union
2. look at sub fields
3. for each subfield, if it's a primitive type, add the mapping to the original typedef. If it's a struct/union, recurse to step 1.

whooo: As most TPMI_RH_ types reflect a TPM2_RH constants, or a list of possible TPM2_RH constants.
By mapping it to TPM2_RH calls like str(ticket.hierarchy) or f"ticket hierachy: {ticket.hierarchy}" will work in an expected way.

whooo:
> If we're generating just prefix it with `constants` or import it `as` so you can have a short form. But since it's generated, i'd just long form it IMO. But i'm open to why we don't want to do that.

I'm not sure I follow you here.
The lgtm-bot alert is due to (in types.py):
```python
import tpm2_pytss.constants as constants                                                                                                                                                                                               
from tpm2_pytss.constants import (                                                                                                                                                                                                     
    TPMA_OBJECT,                                                                                                                                                                                                                       
    TPM2_ALG,                                                                                                                                                                                                                          
    TPM2_ECC_CURVE,                                                                                                                                                                                                                    
)
```
The reason I import constants as that is so I don't have to do a wildcard import but can just check for the type-mapped type in constants, and I think adding constants for all uses of TPMA_OBJECT, TPM2_ALG and TPM2_ECC_CURVE would be meaningless just so get rid of the alert

williamcroberts:
Fix the merge conflict and suppress the warning:
```
If you want to remove only this alert (and not other similar alerts), you can add the comment lgtm [py/import-and-import-from] to the end of line 25. If you make changes to the suppression of alerts in a pull request, please note that theses changes will only be effective once that pull request is merged. 
```

Then let's merge it

whooo:
> Fix the merge conflict and suppress the warning:
> 
> ```
> If you want to remove only this alert (and not other similar alerts), you can add the comment lgtm [py/import-and-import-from] to the end of line 25. If you make changes to the suppression of alerts in a pull request, please note that theses changes will only be effective once that pull request is merged. 
> ```
> 
> Then let's merge it

All done, time for a release candidate?

williamcroberts:
> 
> All done, time for a release candidate?

I want to fix #273 and I have to finish up some paperwork. But otherwise I think we're good.

PR #275, opened 2021-11-30
load_external doc/api update and generic doc updates

williamcroberts:


test/test_types.py
```diff
@@ -309,7 +309,7 @@ def test_TPM2B_PUBLIC(self):
             inPublic.publicArea.parameters.eccDetail.curveID, TPM2_ECC.NIST_P256
         )
 
-    def test_TPM_OBJECT_init(self):
+    def testTPM_OBJECT_init(self):
         pub = TPM2B_PUBLIC(publicArea=TPMT_PUBLIC(nameAlg=TPM2_ALG.SHA256))
         self.assertEqual(pub.publicArea.nameAlg, TPM2_ALG.SHA256)
```
whooo, line 315: Why the rename of the test?

test/test_types.py
```diff
@@ -318,7 +318,7 @@ def test_TPM_OBJECT_init(self):
         ):
             TPM2B_PUBLIC(badfield=1)
 
-    def test_TPM_OBJECT_init_cdata(self):
+    def testTPM_OBJECT_init_cdata(self):
         with self.assertRaises(
             TypeError, msg="Unexpected _cdata type uint8_t, expected TPM2B_DIGEST"
         ):
```
whooo, line 324: Same here

test/test_types.py
williamcroberts: sed got away from me here, will drop.

Issue #203, opened 2021-07-28
Creating EK and AIK


apoorvemohan:
Hi, The current [Python Keylime Agent](https://github.com/keylime/keylime/blob/master/keylime/keylime_agent.py) uses tpm2_tools to create EK (`tpm2_createek`) and AK (`tpm2_createak`), and we want to convert it to start using ESAPI instead. I was wondering if `tpm2-pytss` will provide similar abstractions in the near future? 

Best,
Apoorve

williamcroberts:
tpm2-pytss provides bindings to ESAPI, so you can use it directly and just implement the logic the tools do. I have a ticket open to actually go convert you over to Python ESAPI (Ticket #10) and drop tools, but then I saw you we're heading to RUST. What's the future of the Python code base?

galmasi:
Following up on the comment by @williamcroberts. 

<philosophy> The keylime *agent* is migrating to rust, no question about it. But I don't think I'm ready to dig a shallow grave for the python agent yet -- it's the mainstay of keylime. I believe investing into streamlining the python code to avoid system() calls is a good thing(tm) -- pun not intended, but we'd do the keylime registrar and verifier a service. </philosophy>

<fact> we spent a few hours with @apoorvemohan going through the createEK->createAK->makecredential->activatecredential set of calls, and made them work in pytss. We ended up writing what could end up as templates for RSA based EK and AIK creation. *Do you want them?* if "yes", they are a PR away from pytss. </fact>

williamcroberts:
> Following up on the comment by @williamcroberts.
> 
> The keylime _agent_ is migrating to rust, no question about it. But I don't think I'm ready to dig a shallow grave for the python agent yet -- it's the mainstay of keylime. I believe investing into streamlining the python code to avoid system() calls is a good thing(tm) -- pun not intended, but we'd do the keylime registrar and verifier a service.
> 
> we spent a few hours with @apoorvemohan going through the createEK->createAK->makecredential->activatecredential set of calls, and made them work in pytss. We ended up writing what could end up as templates for RSA based EK and AIK creation. _Do you want them?_ if "yes", they are a PR away from pytss.

Sure we'd love to have this support code so folks can create to tpm2-tool like things effortlessly.

whooo:
@apoorvemohan and @galmasi any updats on this?

williamcroberts:
So what do we need to close this, do we want templates in the python code that can be used in createprimary and create calls?

[...]

PR #279, opened 2021-12-06
utils: add function to generate an EK

whooo:
I have tested the standard ECC and RSA templates towards a real TPM, but the the NV nonce/template is written according to the specification/tpm2-tools so I'm not 100% they work in real life.

Draft until I add types to arguments and documentation.

Related to https://github.com/tpm2-software/tpm2-pytss/issues/203

### DIFF
--- a/tpm2_pytss/utils.py
+++ b/tpm2_pytss/utils.py
@@ -10,10 +10,21 @@ from .internal.crypto import (
     _hmac,
 )
 from .types import *
+from .ESAPI import ESAPI
+from .constants import (
+    ESYS_TR,
+    TPM2_CAP,
+    TPM2_PT_NV,
+    TPM2_ECC,
+    TPM2_PT,
+    TPM2_RH,
+)
+from .internal.templates import _ek
+from .TSS2_Exception import TSS2_Exception
 from cryptography.hazmat.primitives import constant_time as ct
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.backends import default_backend
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Callable
 
 import secrets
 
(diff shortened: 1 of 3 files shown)


tpm2_pytss/utils.py
```diff
@@ -226,3 +227,117 @@ def unwrap(
         )
 
     return s
+
+
+def create_ek(ectx, ekalg=TPM2_ALG.ECC, session=ESYS_TR.NONE):
```
williamcroberts, line 232: I would change this slightly to get_ek_template and be able to pass session1, session2, and session3 here so folks pass everything into the esys calls and then I would have the createprimary call be by the user. This way the session1, which is auth for the ENDORSEMENT hierarchy can be set by the caller to an HMAC session, audit session, etc.

williamcroberts (changes requested):
This is the spec for the EK:
 - https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf

tpm2_pytss/utils.py
williamcroberts: Oh and add type hints to the parameters.

williamcroberts: docstring with link to the generic spec landing page:
  - https://trustedcomputinggroup.org/resource/tcg-ek-credential-profile-for-tpm-family-2-0/

williamcroberts: @idesai could you weigh in on this API design with your thoughts as well?

idesai: I agree. Something to consider... because it uses a TCG EK credential specific value for authorization policy and attributes get_tcg_ek_template might be a good name.

whooo: I kept only one session as the only usefulness I see of passing a session after removing create_primary is for audit and encrypt sessions.
As for the name, I prefer to keep it as close to tpm2_createek as reasonable, to keep it slightly familiar for tpm2-tools users, but I can go either way.

tpm2_pytss/utils.py
```diff
@@ -226,3 +228,129 @@ def unwrap(
...
+        indices.extend(handles)
+
+    if nvtemplate in indices:
+        thandle = ectx.tr_from_tpmpublic(nvtemplate, session1=session)
+        tpub, _ = ectx.nv_read_public(thandle, session1=session)
+        left = tpub.nvPublic.dataSize
+        mtempl = b""
+        while left > 0:
+            off = tpub.nvPublic.dataSize - left
+            size = 0x300 if left > 0x300 else left
```
williamcroberts, line 327: What's this magic number from?

tpm2_pytss/utils.py
```diff
@@ -226,3 +228,129 @@ def unwrap(
...
+
+    nvnonce = nvbase + 1
+    nvtemplate = nvbase + 2
+
+    indices = []
+    more = True
+    while more:
+        more, data = ectx.get_capability(
+            TPM2_CAP.HANDLES,
+            TPM2_HT.NV_INDEX << 24,
```
williamcroberts, line 313: Doesn't this one include the shift: TPM2_HC.HR_NV_INDEX

tpm2_pytss/utils.py
```diff
@@ -226,3 +228,129 @@ def unwrap(
...
+        npub, _ = ectx.nv_read_public(nhandle, session1=session)
+        left = npub.nvPublic.dataSize
+        while left > 0:
+            off = npub.nvPublic.dataSize - left
+            size = 0x300 if left > 0x300 else left
+            data = ectx.nv_read(nhandle, size, off, session2=session)
+            nonce = nonce + bytes(data)
+            left = left - len(data)
+
+    if len(nonce) > 0 and template.publicArea.type == TPM2_ALG.RSA:
```
williamcroberts, line 346: This padding is only done on RSA 2048 and ECC NIST P256 per the 2.2.1.6 "TPMT_PUBLIC Calculation"
  - https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf

Their are other keys possible, so I don't think we want to always do this AFAICT.

williamcroberts (changes requested):

tpm2_pytss/utils.py
williamcroberts: You can have separate audit and encrypt sessions. I was tring to future proof the API a bit since session1,2,3 would just be the defaults and only if folks are interested in that control will they use them.

tpm2_pytss/utils.py
```diff
@@ -226,3 +228,129 @@ def unwrap(
...
+    The template is created according to TCG EK Credential Profile For TPM Family 2.0:
+    - https://trustedcomputinggroup.org/resource/tcg-ek-credential-profile-for-tpm-family-2-0/
+
+    Args:
+        ectx (ESAPI): The ESAPI instance used to look for a possible template and nonce in NV areas
+        ekalg: (TPM2_ALG.ECC or TPM2_ALG.RSA): Type of template to create, defaults to TPM2_ALG.ECC
+        session: (ESYS_TR): Optional session, for encrypt/decrypt and audit sessions, defaults to ESYS_TR.NONE
+
+    Returns:
+        The templase as a TPM2B_PUBLIC instance
```
williamcroberts, line 247: Spelling, templase --> template

tpm2_pytss/utils.py
whooo: I thought that was the minimum allowed for TPM2_PT_NV_BUFFER_MAX, but when I checked again and it's apparently 512, not 768 for the PC client profile

tpm2_pytss/utils.py
whooo: The code currently only supports RSA 2048 and NIST P256 keys (unless the TPM manufacturer has put a non-standard template in one of the NV template areas)

tpm2_pytss/utils.py
whooo: OK, I'll change it to session1 and session2, but I'm worried about session3, the only call that would take all three is get_capability and that would break expectations for session3

tpm2_pytss/utils.py
idesai: > 

This value should ideally be read from the capability structure

williamcroberts:
So the overall design is that it will take a callback to read the NV index and return the template. This way folks can use it with any type of TPM context (ESAPI or FAPI) and a userdata pointer. Since this is python, with closure, we don't really need to worry about passing user data just a callable.

whooo:
> So the overall design is that it will take a callback to read the NV index and return the template. This way folks can use it with any type of TPM context (ESAPI or FAPI) and a userdata pointer. Since this is python, with closure, we don't really need to worry about passing user data just a callable.

So would it be like something like?:
```python
def create_ek_template(ekalg, nv_read_cb):
    defaults_here
    nonce = nv_read_cb(nonce_index)
    template = nv_read_cb(template_index)
    do_rest_here
```

williamcroberts:
> ```python
> ekalg
> ```

Yes, but I forgot to mention ekalg is a string identifier that maps to one of the static templates or one of the NV indices. Let me see if I can figure out what those will be. I don't have them offhand.

williamcroberts:
Here are the strings:

| string | template address | nonce |
|----------|------------------------|---------|
| EK-RSA2048 | 0x01c00003 | 0x01c00004 |
| EK-ECC256 | 0x01c0000b | 0x01c0000c |
| EK-HIGH-RSA2048 | 0x01c00013 | <none> |
| EK-HIGH-ECC256 | 0x01c00015 | <none> |
| EK-HIGH-ECC384 | 0x01c00017 | <none> |
| EK-HIGH-ECC521 | 0x01c00019 | <none> |
| EK-HIGH-ECCSM2P521 | 0x01c0001b | <none> |
| EK-HIGH-RSA3072 | 0x01c0001d | <none> |
| EK-HIGH-RSA4096 | 0x01c0001f | <none> |

This is the callback signature:
```python
def nvread(index)
```
So you could something like:

```python
class MyNVReader:
    def __init__(ectx):
        self.ectx = ectx

    def __call__(self, nvindex):
        return self.ectx.nvread(nvindex)

get_ek("EK-RSA2048", MyNVReader())

# or something like
ectx = ESAPI()
#.....
def my_nv_read(nvindex):
    return ectx.nv_read(nvindex)

get_ek("EK-RSA2048", my_nv_read) 
```

williamcroberts:
If it has a nonce, it needs to call the nv reader cb for the templ and the nonce. The calling framework is responsible for handling merging those and unmarshalling everything into the final TPMT_PUBLIC.

williamcroberts:
I'm not sure what they are going to be for the default static templates. It's not clear to me in the spec if the static templates are the the same as the ones in the NV Index. So these could be mapped to unique strings or based on if callback is None.

See the templates in B.4.4 and later in:
  - https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf

whooo:
is it clear by the specification what should happen if the NV index isn't defined (low template/nonce) on the TPM?
For example should the NV read callback just return an empty buffer in that case?

williamcroberts:
> is it clear by the specification what should happen if the NV index isn't defined (low template/nonce) on the TPM? For example should the NV read callback just return an empty buffer in that case?

Whatever nvread failing for that condition would be. The framework would then use the static nonce.

williamcroberts:
Also look at section "2.2.1.3" in that EK document. It lays out a table with what is and is not allowed, and things that are discouraged.

williamcroberts:
Then the sections on LOW (2.2.1.4) and HIGH (2.2.1.5) layout more of what can be expected. Essentially for the LOW's their SHOULD be an NV index set, however if their isn't it's a default template. For the low range their MAY be a nonce to deal with. For the HIGH range, they're MAY be an NV index for the template, however if not it's the default template.

whooo:
@williamcroberts, when you have time could you check if the code matches the the general idea?
There are still somethings I'm unsure about for the new function specification, for example is it up to the caller or the function to check if a high range certificate NV index is defined or not?

williamcroberts:
> @williamcroberts, when you have time could you check if the code matches the the general idea? There are still somethings I'm unsure about for the new function specification, for example is it up to the caller or the function to check if a high range certificate NV index is defined or not?

The callback function should check. We should probably have a specific case for this error to then go get the default template, all other errors propagate to the original caller. The framework function should never go to the TPM, only through the callback.

whooo:
This is the best I could come up with my understanding to the specification.
I added the certificate to the return values as it must be read in many cases regardless

tpm2_pytss/utils.py
```diff
@@ -226,3 +237,168 @@ def unwrap(
...
+                TPM2_HC.HR_NV_INDEX,
+                4096,
+                session1=session2,
+                session2=session3,
+            )
+            self._indices += list(data.data.handles)
+
+    def __call__(self, index: Union[int, TPM2_RH]) -> bytes:
+        if index not in self._indices:
+            raise NoSuchIndex(index)
```
williamcroberts, line 312: What if the index was defined between the list population and now. Perhaps it's better to just try and read the NV index and look at the error code to see if it's not defined. Which also reduces the trips we need to make to the TPM. Which then, do we even need this exception class, we could just use TSS2 Exceptions that already exist.

tpm2_pytss/utils.py
```diff
@@ -226,3 +237,168 @@ def unwrap(
...
+    """
+
+    def __init__(self, index):
+        self.index = index
+
+    def __str__(self):
+        return f"NV index 0x{index:08x} does not exist"
+
+
+class nv_read_ek:
```
williamcroberts, line 256: Shouldn't classes be capitalized?

tpm2_pytss/utils.py
whooo: I view the chance as low for this use case as new template/certificate areas shouldn't be created during runtime
The problem with letting the TPM throw an error is that it would break audit sessions (and possibly encrypted sessions) for the scenarios where create_ek_template looks for an index but shouldn't fail if there isn't one
We could have a constant of an TSS2_Exception instead, but something explicit and clear is needed, for example someone writes their own callback using FAPI instead which wouldn't throw the same exception

williamcroberts: > I view the chance as low for this use case as new template/certificate areas shouldn't be created during runtime

It's an unlikely race condition, but it's a TOCTOU issue.

> The problem with letting the TPM throw an error is that it would break audit sessions (and possibly encrypted sessions) for 

I'm not following, what exactly breaks? Does a failing command invalidate the session?

> the scenarios where create_ek_template looks for an index but shouldn't fail if there isn't one We could have a constant of an TSS2_Exception instead, but something explicit and clear is needed, for example someone writes their own callback using FAPI instead which wouldn't throw the same exception

True, or if they use their own custom TSS.

whooo: > > I view the chance as low for this use case as new template/certificate areas shouldn't be created during runtime
> 
> It's an unlikely race condition, but it's a TOCTOU issue.
> 
> > The problem with letting the TPM throw an error is that it would break audit sessions (and possibly encrypted sessions) for
> 
> I'm not following, what exactly breaks? Does a failing command invalidate the session?
> 
I reread the specification and I was wrong, so that can be dropped, I'll rework the code to not check the index before trying to read and raise a NoSuchIndex exception if the TSS2_Exception matches the correct error, otherwise reraise the TSS2_Exception

PR #281, opened 2021-12-08
Small fixes

whooo:


tpm2_pytss/constants.py
```diff
@@ -111,6 +111,12 @@ def __str__(self):
             raise ValueError(f"unnmatched values left: 0x{cv:x}")
         return "|".join(ints)
 
+    def __and__(self, value):
+        return self.__class__(int(self) & value)
+
+    def __or__(self, value):
+        return self.__class__(int(self) | value)
```
williamcroberts, line 118: Just bumped into this issue as well. A few things:
  - What about other arithmetic operations? 
  - What about making all things like ESYS_TR have these?

```python
type(ESYS_TR.PCR0 + ESYS_TR.PCR1)
<class 'int'>
```

Right now, any operators change the class back to 'int'.

whooo: I thought about that as well but couldn't figure out an use case, so I skipped it for now.
I can always look into making a int subclass always return the correct type for all the different operators

williamcroberts: > I thought about that as well but couldn't figure out an use case, so I skipped it for now. I can always look into making a int subclass always return the correct type for all the different operators

My gut is telling me to do all the operators for `class TPM_FRIENDLY_INT`, then everyone will pick up "type consistency" across arithmetic operators.

Since they behave like int, I don't see anyone getting suprised, we just want to make sure that both the base class as well as int class works as arguments.

whooo: Right, I'll hack up the other operators (that makes sense) during the week

williamcroberts (approved):

PR #284, opened 2021-12-09
types: add more documentation

williamcroberts:
Document more of the types module and classes and update exception
classes used. ValueError should be used in most instances where
RuntimeError is being used with the exception of the MU lookup.

### DIFF
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -35,6 +35,7 @@ from tpm2_pytss.constants import (
     TPM2_ALG,
     TPM2_ECC_CURVE,
 )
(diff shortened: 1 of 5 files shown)


tpm2_pytss/types.py
```diff
@@ -283,6 +284,8 @@ def __bytes__(self):
         return bytes(buf)
 
     def __str__(self):
+        """foo
+        """
         b = self.__bytes__()
```
whooo, line 289: Seems like this sneaked in

williamcroberts: Which it did, I see the foo. So I have to go update that and resubmit.

PR #290, opened 2021-12-16
replace runtime pkgconfig usage with a version table

whooo:
Currently when using tpm2-tss distribution packages with the bindings
the development package is required during runtime.
This removes the requirement for the pkgconfig files by generating
a version table during the build.


williamcroberts:
Will take post 1.0

williamcroberts (commented):
So my concern on this is that if someone updates the library but doesn't update the bindings, the bindings could try to do something that's incompatible. I really wish we embedded the version into the TSS library to make this easier....

whooo:
Right now the uses for checking the versions seems to be importing new constants or checking for known bugs in FAPI, which shouldn't change anything after a library upgrade, the same risks exists with pkgconfig as well as it doesn't necessarily reflect the dynamic loader search paths

williamcroberts:
> Right now the uses for checking the versions seems to be importing new constants or checking for known bugs in FAPI, which shouldn't change anything after a library upgrade, the same risks exists with pkgconfig as well as it doesn't necessarily reflect the dynamic loader search paths

I guess that's fair, if they update the regular package and not the dev package they'd be in the same scenario.

PR #292, opened 2021-12-16
Fix raises blocks

williamcroberts:


tpm2_pytss/ESAPI.py
```diff
@@ -6809,7 +6808,7 @@ def tr_deserialize(self, buffer: bytes) -> ESYS_TR:
 
         Deserialize the metadata of an ESYS_TR object from a byte buffer that was
         stored on disk for later use by a different program or context.
-        An object can be serialized suing Esys_TR_Serialize.
+        An object can be serialized suing tr_serialize.
```
niooss-ledger, line 6811: "using" is also misspelled "suing" here.

williamcroberts: lolz.

Issue #299, opened 2021-12-21
use-after-free when using sub-objects


niooss-ledger:
Hello,
While writing the issue https://github.com/tpm2-software/tpm2-pytss/issues/298, I stumbled upon a use-after-free issue in tpm2-pytss. For example, let's consider a TPM which has a Storage Root Key at `0x81000000`, which was for example created using:

```sh
tpm2_createprimary -c /tmp/prim -C o
tpm2_evictcontrol -c /tmp/prim -o 0x81000000
```

`tpm2_readpublic -c 0x81000000` shows that this key is of type `0x1` (RSA) with attributes `0x30072` (`fixedtpm|fixedparent|sensitivedataorigin|userwithauth|restricted|decrypt`). And it is possible to use tpm2-pytss to see this:

```python
from tpm2_pytss import *
ectx = ESAPI()
object_handle = ectx.tr_from_tpmpublic(0x81000000)
res = ectx.read_public(object_handle)
print(int(res[0].publicArea.type))  # "1"
print(res[0].publicArea.type)  # "rsa"
print(hex(res[0].publicArea.objectAttributes)) # "0x30072"
```

But with a slightly different code, the values become random:

```python
from tpm2_pytss import *
ectx = ESAPI()
object_handle = ectx.tr_from_tpmpublic(0x81000000)
pub = ectx.read_public(object_handle)[0].publicArea
print(int(pub.type))  # "0"
print(pub.type)  # "error"
print(hex(pub.objectAttributes)) # "0x1e2d010"
```

To better understand what is going on, I instrumented `read_public` to show memory allocation and freeing of the `TPM2B_PUBLIC` structure:

```diff
diff --git a/tpm2_pytss/ESAPI.py b/tpm2_pytss/ESAPI.py
--- a/tpm2_pytss/ESAPI.py
+++ b/tpm2_pytss/ESAPI.py
@@ -950,8 +950,12 @@ class ESAPI:
                 qualified_name,
             )
         )
+        print(f"ESAPI.read_public: out_public[0] = {out_public[0]}")
+        def print_and_free(p):
+            print(f"ESAPI.read_public: (gc) free {p}")
+            return lib.Esys_Free(p)
         return (
-            TPM2B_PUBLIC(_get_dptr(out_public, lib.Esys_Free)),
+            TPM2B_PUBLIC(_get_dptr(out_public, print_and_free)),
             TPM2B_NAME(_get_dptr(name, lib.Esys_Free)),
[...]

PR #300, opened 2021-12-23
types: ensure deconstructor order

whooo:
When using one of the tpm2-tss deconstructors (Esys_Free and Fapi_Free)
the memory might be freed when there is no more reference to the root
field even when a sub-field is still used, leading to memory corruption.

This solves this by using a custom deconstructor for sub-fields which
keeps a reference to its parent

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/299

### DIFF
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-2
 
 import unittest
+import gc
 
 from tpm2_pytss import *
 from .TSS2_BaseTest import TSS2_EsapiTest
@@ -4687,6 +4688,33 @@ class TestEsys(TSS2_EsapiTest):
 
         with self.assertRaises(TSS2_Exception):
             self.ectx.tr_deserialize(b"0123456890")
+
+    def test_ref_parent(self):
+        # Test keeping a reference to the parent in a child structure
+        _, pub, _, _, _ = self.ectx.create_primary(None, "ecc")
+        pa = pub.publicArea
+        del pub
+        gc.collect()
+        self.assertEqual(pa.type, TPM2_ALG.ECC)
+        self.assertEqual(
+            pa.objectAttributes, TPMA_OBJECT.DEFAULT_TPM2_TOOLS_CREATEPRIMARY_ATTRS
+        )
+
+        # Test keeping a reference to the parent in an element from a TPML_OBJECT
+        _, sels, _ = self.ectx.pcr_read("sha256:16")
+        sel = sels[0]
+        del sels
+        gc.collect()
+        self.assertEqual(sel.hash, TPM2_ALG.SHA256)
+        self.assertEqual(bytes(sel.pcrSelect), b"\x00\x00\x01\x00")
+
+        # Test keeping a reference to the parent when accessing a TPM2_SIMPLE_OBJECT buffer
+        handle, _, _, _, _ = self.ectx.create_primary(None, "ecc")
+        _, name, _ = self.ectx.read_public(handle)
+        algb = name.name[0:2]
+        del name
+        gc.collect()
+        self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
 
 
 if __name__ == "__main__":
--- a/tpm2_pytss/internal/utils.py
+++ b/tpm2_pytss/internal/utils.py
@@ -125,10 +125,24 @@ def _fixup_cdata_kwargs(this, _cdata, kwargs):
     return (_cdata, kwargs)
 
 
-def _convert_to_python_native(global_map, data):
+def _ref_parent(data, parent):
+    tipe = ffi.typeof(parent)
+    if tipe.kind != "pointer":
+        return data
+
+    def deconstructor(ptr):
+        parent
+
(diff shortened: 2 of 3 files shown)


test/test_esapi.py
```diff
@@ -4688,6 +4688,21 @@ def test_esys_tr_functions(self):
...
+            pub.objectAttributes, TPMA_OBJECT.DEFAULT_TPM2_TOOLS_CREATEPRIMARY_ATTRS
+        )
+
+        sel = self.ectx.pcr_read("sha256:16")[1][0]
+        self.assertEqual(sel.hash, TPM2_ALG.SHA256)
+        self.assertEqual(bytes(sel.pcrSelect), b"\x00\x00\x01\x00")
+
+        handle = self.ectx.create_primary(None, "ecc")[0]
+        algb = self.ectx.read_public(handle)[1][0:2]
+        self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
```
williamcroberts, line 4704: How does this test it? Shouldn't we force a GC event? Ie `gc.collect()`

williamcroberts:
Can you rebase this on master, I have the build break fixed.

williamcroberts:
I'm thinking more along this for the test, but otherwise LGTM. Interesting way to use that ffi.gc call, any reasons why you took that approach over say embedding the cdata reference in the subordinate object?

```diff
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3 -u
 # SPDX-License-Identifier: BSD-2
 
+import gc
 import unittest
 
 from tpm2_pytss import *
@@ -4688,19 +4689,20 @@ class TestEsys(TSS2_EsapiTest):
         with self.assertRaises(TSS2_Exception):
             self.ectx.tr_deserialize(b"0123456890")
 
-    def test_ref_parent(self):
-        pub = self.ectx.create_primary(None, "ecc")[1].publicArea
-        self.assertEqual(pub.type, TPM2_ALG.ECC)
-        self.assertEqual(
-            pub.objectAttributes, TPMA_OBJECT.DEFAULT_TPM2_TOOLS_CREATEPRIMARY_ATTRS
-        )
+    def test_sub_object_cdata_validity_no_parent(self):
 
+        # get some sub-parent objects with no references to the parents.
+        # and ensure a GC event occurs to reap non-referenced parents
+        # and verify that the cdata in the sub object is still valid.
         sel = self.ectx.pcr_read("sha256:16")[1][0]
-        self.assertEqual(sel.hash, TPM2_ALG.SHA256)
-        self.assertEqual(bytes(sel.pcrSelect), b"\x00\x00\x01\x00")
 
         handle = self.ectx.create_primary(None, "ecc")[0]
         algb = self.ectx.read_public(handle)[1][0:2]
+
+        gc.collect()
+
+        self.assertEqual(sel.hash, TPM2_ALG.SHA256)
+        self.assertEqual(bytes(sel.pcrSelect), b"\x00\x00\x01\x00")
         self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
```

whooo:
> I'm thinking more along this for the test, but otherwise LGTM. Interesting way to use that ffi.gc call, any reasons why you took that approach over say embedding the cdata reference in the subordinate object?
> 
I'll look into forcing a garbage collection in the test to hardened them (and make them more readable).
IIRC the reason I went with ffi.gc instead of an attribute was that some types aren't to happy with an unknown attribute being set for them

williamcroberts:
> > I'm thinking more along this for the test, but otherwise LGTM. Interesting way to use that ffi.gc call, any reasons why you took that approach over say embedding the cdata reference in the subordinate object?
> 
> I'll look into forcing a garbage collection in the test to hardened them (and make them more readable). IIRC the reason I went with ffi.gc instead of an attribute was that some types aren't to happy with an unknown attribute being set for them

Do you remember what the explicit error was? Python just seems to take it via setattr.... perhaps it was all of our wrapping logic and we could have just added it to TPM_OBJECT class?

```python
>>> class FOO(object):
...     pass
... 
>>> f = FOO()
>>> setattr(f, "bill", "value")
>>> print(f)
<__main__.FOO object at 0x7efebbcaa2b0>
>>> print(f.bill)
value
```

whooo:
> > > I'm thinking more along this for the test, but otherwise LGTM. Interesting way to use that ffi.gc call, any reasons why you took that approach over say embedding the cdata reference in the subordinate object?
> > 
> > 
> > I'll look into forcing a garbage collection in the test to hardened them (and make them more readable). IIRC the reason I went with ffi.gc instead of an attribute was that some types aren't to happy with an unknown attribute being set for them
> 
> Do you remember what the explicit error was? Python just seems to take it via setattr.... perhaps it was all of our wrapping logic and we could have just added it to TPM_OBJECT class?
> 
I think it was AttributeError or something similar, the issue was with types that aren't being wrapped by TPM_OBJECT, for example pctSelect in TPMS_PCR_SELECTION or when accessing the buffer in a TPM2B_DIGEST instance.

I changed the code a bit so a reference to the parent is only kept if the parent is a pointer which should reduce the overhead a bit.

williamcroberts (approved):

PR #301, opened 2022-01-04
types: add check for sphinx in type_mapping import

whooo:
Any errors when trying to import type_mapping should only be ignored during doc generation.

### DIFF
--- a/tpm2_pytss/types.py
+++ b/tpm2_pytss/types.py
@@ -36,12 +36,14 @@ from tpm2_pytss.constants import (
     TPM2_ECC_CURVE,
 )
 from typing import Union, Tuple
+import sys
 
 try:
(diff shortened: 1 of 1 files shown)


williamcroberts (approved):

williamcroberts:
@whooo you ready for an RC1? If so I'll roll one.

whooo:
> @whooo you ready for an RC1? If so I'll roll one.

I'll write a fix for https://github.com/tpm2-software/tpm2-pytss/issues/305
Other then that I'm ready for RC1

williamcroberts:
> > @whooo you ready for an RC1? If so I'll roll one.
> 
> I'll write a fix for #305 Other then that I'm ready for RC1

Will do RC1 on Jan 10

Issue #297, opened 2021-12-21
ESAPI: why do NV functions use "auth_handle: ESYS_TR = 0"?


niooss-ledger:
Hello,
In tpm2-pytss 1.0.0-rc0, some functions in ESAPI take a parameter `auth_handle: ESYS_TR = 0` (mostly the ones related to NV index). For example https://github.com/tpm2-software/tpm2-pytss/blob/7cd4cc0632f87b51483dc59630e7fb2d6abd2630/tpm2_pytss/ESAPI.py#L6170-L6179

This zero means "use the handle from `nv_index`", and this is implemented for example in https://github.com/tpm2-software/tpm2-pytss/blob/7cd4cc0632f87b51483dc59630e7fb2d6abd2630/tpm2_pytss/ESAPI.py#L6205-L6206

Usually in Python, default values use `None` instead. Moreover here, zero could have a special meaning: the enum `ESYS_TR` contains an entry for `0`, which is `ESYS_TR.PCR0`:

```python
>>> ESYS_TR.PCR0
0
```

It seems that these `auth_handle=0` were introduced in commit https://github.com/tpm2-software/tpm2-pytss/commit/40ec6dc1093ed3cb16bcc42f1e9e5c88829aef48 and documented by:

```text
For basic NV configurations work. Also, modify the arguments to NV_Read
and NV_Write so we can just default the authHandle to the nvIndex handle
if the authHandle is not specified. This makes calls simpler in the more
common auth to the handle role.
```

In the end, my questions are: why was zero used, instead of `None`? Can this be modified (this will help checking the type annotations of `ESAPI.py`)?

williamcroberts:
Now is a great time to modify that before we actually fully roll 1.0. I have no objections to using None over 0. I chose 0 since it's unused in NV index space, but I think None is a better choice.

PR #304, opened 2022-01-04
ESAPI: use None instead of 0 for default auth_handle

niooss-ledger:
Several functions related to NV indexes take an optional parameter defined as `auth_handle: ESYS_TR = 0`. However the zero `ESYS_TR` is defined (it is `ESYS_TR.PCR0`) and in Python, such "default parameter value" usually uses `None`. Moreover using `None` helps checking the type annotations.

Make the functions use `None` instead of `0`.

Fixes: https://github.com/tpm2-software/tpm2-pytss/issues/297


williamcroberts (approved):
Lets pull this in for 1.0 since it would be an API change later

PR #309, opened 2022-01-05
Fix set auth naming

williamcroberts:


niooss-ledger (approved):

tpm2_pytss/ESAPI.py
```diff
@@ -245,7 +245,7 @@ def tr_close(self, esys_handle: ESYS_TR):
         esys_tr_ptr[0] = esys_handle
         _chkrc(lib.Esys_TR_Close(self._ctx, esys_tr_ptr))
 
-    def set_auth(
+    def tr_set_auth(
         self, esys_handle: ESYS_TR, auth_value: Union[TPM2B_AUTH, bytes, str, None]
     ):
         """Set the authorization value of an ESYS_TR.
```
whooo, line 251: The docstring has a reference to set_auth, "Therefore set_auth needs to be called again after" which probably should be updated

whooo (changes requested):

PR #310, opened 2022-01-05
ESAPI: fix double-close issue

niooss-ledger:
When `ESAPI.close` is called twice, it could raise an exception:

    >>> from tpm2_pytss import *
    >>> ectx = ESAPI("tabrmd:bus_type=system")
    >>> ectx.close()
    >>> ectx.close()
    WARNING:esys:src/tss2-esys/esys_context.c:114:Esys_Finalize() Finalizing NULL context.
      File "/workspace/tpm2-pytss/tpm2_pytss/ESAPI.py", line 155, in close
        self._tcti.close()

Fix this by ensuring that `self._tcti` is not `None` before calling `close`. While at it, do not call `Esys_Finalize` twice.

### DIFF
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -4716,6 +4716,21 @@ class TestEsys(TSS2_EsapiTest):
         gc.collect()
         self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
 
+    def test_double_close(self):
+        ectx = ESAPI(self.tpm.tcti_name_conf)
+        self.assertTrue(ectx._did_load_tcti)
+        self.assertTrue(ectx._ctx)
+        self.assertTrue(ectx._ctx_pp)
+        self.assertEqual(ectx.tcti.name_conf, self.tpm.tcti_name_conf)
+        ectx.close()
+        self.assertFalse(ectx._ctx)
+        self.assertFalse(ectx._ctx_pp)
+        self.assertEqual(ectx.tcti, None)
+        ectx.close()
+        self.assertFalse(ectx._ctx)
+        self.assertFalse(ectx._ctx_pp)
+        self.assertEqual(ectx.tcti, None)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tpm2_pytss/ESAPI.py
+++ b/tpm2_pytss/ESAPI.py
@@ -148,10 +148,11 @@ class ESAPI:
 
         C Function: Esys_Finalize
         """
-
-        lib.Esys_Finalize(self._ctx_pp)
-        self._ctx = ffi.NULL
-        if self._did_load_tcti:
+        if self._ctx_pp:
+            lib.Esys_Finalize(self._ctx_pp)
+            self._ctx = ffi.NULL
+            self._ctx_pp = ffi.NULL
+        if self._did_load_tcti and self._tcti is not None:
             self._tcti.close()
         self._tcti = None
 


test/test_esapi.py
```diff
@@ -4716,6 +4716,21 @@ def test_ref_parent(self):
...
+        self.assertTrue(bool(ectx._ctx))
+        self.assertTrue(bool(ectx._ctx_pp))
+        self.assertEqual(ectx.tcti.name_conf, self.tpm.tcti_name_conf)
+        ectx.close()
+        self.assertFalse(bool(ectx._ctx))
+        self.assertFalse(bool(ectx._ctx_pp))
+        self.assertEqual(ectx.tcti, None)
+        ectx.close()
+        self.assertFalse(bool(ectx._ctx))
+        self.assertFalse(bool(ectx._ctx_pp))
```
whooo, line 4731: You could probably skip bool(...) for assertTrue and assertFalse calls

niooss-ledger: I agree. I was not familiar with `assertTrue`/`assertFalse` but the documentation is clear on https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue : `assertTrue(expr)` checks that `bool(expr) is True`, not that `expr is True`.
Thanks for the review!

whooo:
I'm restarted the jobs, with it's weird that all that timed out was using tpm2-tss 2.4.x, so might be a bug somewhere

williamcroberts:
This breaks something, an empty commit here causes all the runners to pass:
  - https://github.com/tpm2-software/tpm2-pytss/pull/314

test/test_esapi.py
```diff
@@ -4716,6 +4716,21 @@ def test_ref_parent(self):
         gc.collect()
         self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
 
+    def test_double_close(self):
```
williamcroberts, line 4719: This test seems to hang for me locally as well.

williamcroberts: Oh I think I see it. We have a connection to mssim in self.ectx, and use the same information to try and connect to it again, but only one client can be connected at a time.

williamcroberts:
I found the bug, this patch fixes it. Two things:
1. We need to close the self.ectx to create our own connection without being blocked to the simulator.
2. We need to always close the TCTI when a tcti is passed to ESAPI. The internal library will not manage it.


```diff
diff --git a/test/test_esapi.py b/test/test_esapi.py
index 972bdb11a1c8..bd6c4d9a398b 100644
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -4717,6 +4717,7 @@ class TestEsys(TSS2_EsapiTest):
         self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
 
     def test_double_close(self):
+        self.ectx.close()
         ectx = ESAPI(self.tpm.tcti_name_conf)
         self.assertTrue(ectx._did_load_tcti)
         self.assertTrue(ectx._ctx)
diff --git a/tpm2_pytss/ESAPI.py b/tpm2_pytss/ESAPI.py
index ac7f75373d8a..0df023159a59 100644
--- a/tpm2_pytss/ESAPI.py
+++ b/tpm2_pytss/ESAPI.py
@@ -111,11 +111,10 @@ class ESAPI:
                 f"Expected tcti to be type TCTI, str or None, got {type(tcti)}"
             )
 
-        self._passed_tcti = self._did_load_tcti = False
+        self._passed_tcti = False
 
         # support tpm2-tools style tcti strings
         if isinstance(tcti, str):
-            self._did_load_tcti = True
             self._passed_tcti = True
             tcti = TCTILdr.parse(tcti)
         elif isinstance(tcti, TCTI):
@@ -152,7 +151,7 @@ class ESAPI:
             lib.Esys_Finalize(self._ctx_pp)
             self._ctx = ffi.NULL
             self._ctx_pp = ffi.NULL
-        if self._did_load_tcti and self._tcti is not None:
+        if self._passed_tcti and self._tcti is not None:
             self._tcti.close()
         self._tcti = None
```

williamcroberts:
I applied this patch and have fixed it over here:
  - https://github.com/tpm2-software/tpm2-pytss/compare/master...williamcroberts:fix-double-esapi-close?expand=1

If that passes, I will merge and close this one.

williamcroberts:
Taking some adjustments to this over on my branch, so closing this:
  - https://github.com/tpm2-software/tpm2-pytss/pull/315

Issue #298, opened 2021-12-21
[Feedback] Function tr_from_tpmpublic is not easy enough to discover


niooss-ledger:
Hello,

Here is some feedback about what it felt like to use tpm2-pytss 1.0.0-rc0 to perform operations such as reading a SRK or a NV index. Let's start with trying to read a Storage Root Key stored at `0x81000000`. In a shell, this is easy:

```console
$ tpm2_readpublic -c 0x81000000
...
name-alg:
  value: sha256
  raw: 0xb
attributes:
  value: fixedtpm|fixedparent|sensitivedataorigin|userwithauth|restricted|decrypt
  raw: 0x30072
type:
  value: rsa
  raw: 0x1
exponent: 65537
bits: 2048
...
```

But with tpm2-pytss, this is not as straightforward:

```python
>>> from tpm2_pytss import *
>>> ectx = ESAPI()
>>> ectx.read_public(0x81000000)
Traceback (most recent call last):
...
TypeError: expected object_handle to be type ESYS_TR, got <class 'int'>

>>> ectx.read_public(ESYS_TR(0x81000000))
ERROR:esys:src/tss2-esys/esys_iutil.c:1095:esys_GetResourceObject() Error: Esys handle does not exist (70018). 
ERROR:esys:src/tss2-esys/api/Esys_ReadPublic.c:170:Esys_ReadPublic_Async() objectHandle unknown. ErrorCode (0x00070018) 
ERROR:esys:src/tss2-esys/api/Esys_ReadPublic.c:81:Esys_ReadPublic() Error in async function ErrorCode (0x00070018) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/git/tpm2_pytss/ESAPI.py", line 933, in read_public
    _chkrc(
  File "/git/tpm2_pytss/internal/utils.py", line 24, in _chkrc
    raise TSS2_Exception(rc)
tpm2_pytss.TSS2_Exception.TSS2_Exception: esapi:The ESYS_TR resource object is bad
```

Reading the documentation of `read_public` does not help. And it is actually wrong! https://github.com/tpm2-software/tpm2-pytss/blob/f79d3be51a0c74ac0cd60c7fe34dcb55b7520197/tpm2_pytss/ESAPI.py#L888-L907

This documentation describes args `in_private` and `in_public`, which do not exist, and does not describe `object_handle`.

[...]

PR #311, opened 2022-01-06
test_esapi: use tr_from_tpmpublic with NV index

niooss-ledger:
In order to read a NV index from a "TPM index", the function `tr_from_tpmpublic` needs to be used first. Add such a use-case in the test suite.

This was discussed in https://github.com/tpm2-software/tpm2-pytss/issues/298.


whooo:
FYI, I'm currently working on some examples, one which covers this (and our discussion in the issue), so I'm not really sure this is needed as it doesn't increase the coverage.
I'll probably push a draft of the examples in a couple of days

williamcroberts:
Seems to be addressed in: 8fd7d70b8

PR #312, opened 2022-01-06
ci: define DEBIAN_FRONTEND correctly in apt-get

niooss-ledger:
When using `sudo apt-get ...`, the environment variables are not forwarded by default. So in fact `apt-get` runs without
`DEBIAN_FRONTEND=noninteractive`. This triggers interactive prompts when running `.ci/install-deps.sh` in a Docker container.

Fix this by defining `DEBIAN_FRONTEND=noninteractive` directly in the `sudo` line.

### DIFF
--- a/.ci/install-deps.sh
+++ b/.ci/install-deps.sh
@@ -8,9 +8,8 @@ export TPM2_TSS_VERSION=${TPM2_TSS_VERSION:-"3.0.3"}
 #
 # Get dependencies for building and install tpm2-tss and abrmd projects
 #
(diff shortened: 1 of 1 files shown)


williamcroberts:
I think the RTD issue is related to this:
  - https://stackoverflow.com/questions/70602290/google-app-engine-deployment-fails-error-while-finding-module-specification-for

Looks like that release has been yanked.

williamcroberts:
Retriggered docs, passing will merge. Is this why https://github.com/tpm2-software/tpm2-pytss/pull/310 is stuck?

PR #313, opened 2022-01-07
constants: add fixes for tools-like string generation

whooo:
Fixes:
Make TPMA_CC a subclass of TPMA_FRIENDLY_INTLIST
str(TPMA_CC()) returning "commandindex_shift"
str(TPMA_NV.AUTHWRITE | ...) raising ValueError: unnmatched values left: 0x4


whooo:
@williamcroberts, I think this should go into the next RC

williamcroberts (approved):

PR #315, opened 2022-01-07
ESAPI: fix double-close issue

williamcroberts:
When `ESAPI.close` is called twice, it could raise an exception:

    >>> from tpm2_pytss import *
    >>> ectx = ESAPI("tabrmd:bus_type=system")
    >>> ectx.close()
    >>> ectx.close()
    WARNING:esys:src/tss2-esys/esys_context.c:114:Esys_Finalize() Finalizing NULL context.
      File "/workspace/tpm2-pytss/tpm2_pytss/ESAPI.py", line 155, in close
        self._tcti.close()

Fix this by ensuring that `self._tcti` is not `None` before calling
`close`. While at it, do not call `Esys_Finalize` twice.

Fix a bug in the handling of the logic to close the TCTI in
ESAPI.close(). If a TCTI object is passed to ESAPI(), the close() method
on the TCTI is never invoked and the connection is never terminated,
thus subsequent calls will block as the TPM resource is still held.

Signed-off-by: Nicolas Iooss <nicolas.iooss@ledger.fr>

### DIFF
--- a/test/test_esapi.py
+++ b/test/test_esapi.py
@@ -4716,6 +4716,29 @@ class TestEsys(TSS2_EsapiTest):
         gc.collect()
         self.assertEqual(bytes(algb), TPM2_ALG.SHA256.to_bytes(2, "big"))
 
+    def test_double_close(self):
+
+        # Shutdown the old TCTI connection so we can connect with a name-conf string
+        # without blocking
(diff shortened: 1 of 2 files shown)


williamcroberts:
Ahh FAPI broke, I see the issue, fixing.

PR #317, opened 2022-01-12
examples: create examples structure and add first example

whooo:
This is a starting point for example usage of the different APIs, starting with a basic example of using ESAPI.

My plan is to add more examples over time which somewhat should cover real world use cases while being readable.

### DIFF
--- a/examples/tpm2_filternv/README.md
+++ b/examples/tpm2_filternv/README.md
@@ -1,0 +1,17 @@
+# tpm2_nvfilter
+
+Display a list of NV indices which matches zero or more attributes, such as `written` or `authread`.
+```console
+$ tpm2_nvfilter --filter "platformcreate|policy_delete|writelocked"
+0x1800001:
+  name: 000b70f1fe7cb8045d9ff1ef39fc64cf247cecad6e1864fb586badab244cac7dfd6a
+  hash algorithm:
+    friendly: sha256
+    value: 0xb
+  attributes:
+    friendly: authwrite|policy_delete|writelocked|writedefine|authread|noda|written|platformcreate
+    value: 0x62042c04
+  size: 70
+  authorization policy: 1169A46A813A8CCDD0F3066785207BB9B67AFD3A6CD6DFE5C5AEE120867A96DF
+
+```
--- a/examples/tpm2_filternv/tpm2_filternv
+++ b/examples/tpm2_filternv/tpm2_filternv
@@ -1,0 +1,75 @@
+#!/usr/bin/python3
+
+import argparse
+import os
+from binascii import hexlify
+from tpm2_pytss.ESAPI import ESAPI
+from tpm2_pytss.TCTILdr import TCTILdr
+from tpm2_pytss.constants import TPMA_NV, TPM2_CAP, TPM2_HC
+
+
+output_template = """{pub.nvIndex:#x}:
+  name: {hexname}
+  hash algorithm:
+    friendly: {pub.nameAlg}
+    value: {pub.nameAlg:#x}
+  attributes:
+    friendly: {pub.attributes}
+    value: {pub.attributes:#x}
+  size: {pub.dataSize}
+"""
+
+
+def get_tcti(arg):
+    if arg:
+        return TCTILdr.parse(arg)
+    elif os.getenv("PYTSS_EXAMPLE_TCTI"):
+        return TCTILdr.parse(os.getenv("PYTSS_EXAMPLE_TCTI"))
+    return TCTILdr()
+
+
+def tpma_nv_arg(arg):
+    try:
+        return TPMA_NV(arg, base=0)
+    except ValueError:
+        pass
+    return TPMA_NV.parse(arg)
+
+
+def run(ectx, filter_attrs):
+    indices = list()
+    more = True
+    while more:
+        more, data = ectx.get_capability(TPM2_CAP.HANDLES, TPM2_HC.HR_NV_INDEX, 4096)
+        indices += list(data.data.handles)
+
+    for index in indices:
+        handle = ectx.tr_from_tpmpublic(index)
+        pub2b, name = ectx.nv_read_public(handle)
+        pub = pub2b.nvPublic
+        if pub.attributes & filter_attrs != filter_attrs:
+            continue
+        hexname = hexlify(name.name).decode("ascii")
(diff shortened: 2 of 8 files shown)


whooo:
@niooss-ledger, could you have a look and see if you think this can close https://github.com/tpm2-software/tpm2-pytss/issues/298

williamcroberts:
I wonder if we should mention in the ESYS_TR class documentation that to go from a TPM2_HANDLE to an ESYS_TR you should use tr_from_public.

whooo:
> I wonder if we should mention in the ESYS_TR class documentation that to go from a TPM2_HANDLE to an ESYS_TR you should use tr_from_public.

Wouldn't hurt, I can create another PR for that.

examples/tpm2_filternv/README.md
```diff
@@ -0,0 +1,17 @@
+# tpm2_nvfilter
+
+Display a list of NV indices which matches zero or more attributes, such as `written` or `authread`.
+```
```
niooss-ledger, line 4: You could use ` ```console` to format this snippet as an interactive shell session.

examples/tpm2_filternv/tpm2_filternv
```diff
@@ -0,0 +1,71 @@
...
+def get_tcti(arg):
+    if arg:
+        return TCTILdr.parse(arg)
+    elif os.getenv("PYTSS_EXAMPLE_TCTI"):
+        return TCTILdr.parse(os.getenv("PYTSS_EXAMPLE_TCTI"))
+    return TCTILdr()
+
+
+def tpma_nv_arg(arg):
+    return TPMA_NV.parse(arg)
```
niooss-ledger, line 32: (Suggestion) This could accept integers too:

```python
def tpma_nv_arg(arg):
    try:
        return TPMA_NV.parse(arg)
    except ValueError:
        # Decode an integer
        return TPMA_NV(int(arg, 0))
```

This way, using `./tpm2_filternv -F 0x10` could work, for example. What do you think?
(Or `parse` could be extended to support integers such as `"1"` and `"0x10"`)

niooss-ledger (commented):
Thanks for this PR! I agree this would be a nice way to close https://github.com/tpm2-software/tpm2-pytss/issues/298 :)

I added some comments. Also, the example does not work as-is on a TPM I am using, due to an issue that I reported in https://github.com/tpm2-software/tpm2-pytss/issues/318, but when I work around this issue, I managed to run it. And the code of the example is easy to understand :)

examples/tpm2_filternv/README.md
whooo: Ah, missed that console exists, it's added now

examples/tpm2_filternv/tpm2_filternv
whooo: Good idea, added a variation of that

examples/tpm2_filternv/tpm2_filternv
```diff
@@ -0,0 +1,75 @@
...
+        return TCTILdr.parse(arg)
+    elif os.getenv("PYTSS_EXAMPLE_TCTI"):
+        return TCTILdr.parse(os.getenv("PYTSS_EXAMPLE_TCTI"))
+    return TCTILdr()
+
+
+def tpma_nv_arg(arg):
+    try:
+        return TPMA_NV(arg, 0)
+    except:
```
niooss-ledger, line 34: In some other projects, I saw a generic Python recommendation of avoiding the use of bare `except:` statements. Here it seems that the only exception which is expected to happen is `ValueError`. Is it possible to use `except ValueError:` instead?

By the way, the `TPMA_NV(arg, 0)` syntax is difficult to understand for newcomers who can wonder: what is this `0`? Is this syntax documented somewhere? In my humble opinion it is clearer to use `TPMA_NV(int(arg, 0))`, or if you want to about this call to `int()`, something like `TPMA_NV(arg, base=0)` to hint that the constructor of `int` is invoked here. What do you think?

williamcroberts:
I have mixed feelings about example code as well, it's often times left to rot, plus how do we know if something we do breaks an example. We need to figure out a way to get this into our test suite.

examples/tpm2_filternv/tpm2_filternv
whooo: good points, fixed

whooo:
> I have mixed feelings about example code as well, it's often times left to rot, plus how do we know if something we do breaks an example. We need to figure out a way to get this into our test suite.

I worry about the rot as well and I think having good tests might be hard as some of the examples I plan on writing will require extra dependencies.
I'm open to having the examples in another git repository (for example directly under my user on github) with a big warning about testing/compatibility

williamcroberts:
> > I have mixed feelings about example code as well, it's often times left to rot, plus how do we know if something we do breaks an example. We need to figure out a way to get this into our test suite.
> 
> I worry about the rot as well and I think having good tests might be hard as some of the examples I plan on writing will require extra dependencies. I'm open to having the examples in another git repository (for example directly under my user on github) with a big warning about testing/compatibility

Wouldn't they just be dev dependencies? As long as you can pip install and add them to the dev section, I don't really care.
We've also talked about having more tutorials and stuff under our github io page, so that's an option too, or we could embed them in the docs as a tutorials section for just pytss. I'm pretty open to ideas here, and not really invested in any one way.

whooo:
> > I worry about the rot as well and I think having good tests might be hard as some of the examples I plan on writing will require extra dependencies. I'm open to having the examples in another git repository (for example directly under my user on github) with a big warning about testing/compatibility
> 
> Wouldn't they just be dev dependencies? As long as you can pip install and add them to the dev section, I don't really care. We've also talked about having more tutorials and stuff under our github io page, so that's an option too, or we could embed them in the docs as a tutorials section for just pytss. I'm pretty open to ideas here, and not really invested in any one way.

Putting the dependencies under dev should be enough for now.
I added a test for the first example, I didn't add full coverage as I don't think it's needed but I might be wrong

iolivergithub:
FYI, I've started putting examples up at: https://github.com/tpm2dev/tpm.dev.tutorials/tree/master/PythonExamples
Feel free to use

williamcroberts:
I think if we have examples, to ensure they are complete they should run with the test harness, i'm sure we could plumb them in somehow.

whooo:
I'll close this for now, might revisit it in the future

PR #326, opened 2022-02-09
types: add check in TPMS_CONTEXT.to_tools for session handles

whooo:
Instead of letting the caller keep track of if they passed session_type and
auth_hash to to_tools to get a tpm2-tools session context object raise an
exception if session_type and auth_hash is missing when  savedHandle is is
in the HMAC or policy session ranges.


whooo:
@williamcroberts, could you have a look at this, I think this should go into the next RC

PR #327, opened 2022-02-10
encoding: add tpm2-tools encoder

whooo:
TODO:
- [x] tpm2_getcap properties-fixed
- [x] tpm2_getcap properties-variable
- [x] tpm2_getcap ecc-curves
- [x] tpm2_getcap handles-*
- [X] tpm2_readpublic
- [X] tpm2_print
- [x] tpm2_nvreadpublic
- [x] ~~tpm2_quote~~
- [x] tpm2_createak
- [x] ~~tpm2_getekcertificate~~
- [x] tpm2_getcap PCRs
- [x] tpm2_pcrread
- [x] tpm2_incrementalselftest
- [x] tpm2_load
- [x] tpm2_loadexternal
- [x] tpm2_pcrevent
- [x] tpm2_pcrallocate
- [x] ~~tpm2_send (may be in verbose mode?)~~
- [x] tpm2_sessionconfig
- [x] ~~switch to OrderedDict~~
- [x] Go through all the tools to check for different formats
- [x] Add decoding support
- [x] Block unsupported TPML types
- [x] Do visual inspection with actual YAML output from the encoder
- [x] ~~Add name when encoding TPMT_PUBLIC~~
- [x] Fix type hints
- [x] Add to_yaml and from_yaml

Open questions:
* Should we encode it as YAML for the user or let them handle that part?
* If we should do it, must it be character by character compatible? (See indentation in tpm2_getcap algorithms for example)  

Known differences to tpm2-tools:
tpm2_readpublic has the following output for when scheme-halg is TPM2_ALG_ERROR:
```yaml
scheme-halg:
  value: (null)
  raw: 0x0
```
Which might be a bug
Known issues:
* It's not always possible to decode a TPMS_ATTEST yaml output from tpm2_print
* Some data is lost for some of the TPMA types (only for reserved values, so not a big issue)


williamcroberts:
<snip>
> Open questions:
> 
> * Should we encode it as YAML for the user or let them handle that part?

Yes return YAML. How else would you do it out of curiosity?

> * If we should do it, must it be character by character compatible? (See indentation in tpm2_getcap algorithms for example)

No, ordering/whitespace etc doesn't mater. Essentially if you load one document from tools and the other from python, as long as each key/value pair exists that's good enough, and nothing extra is hanging out.

Looking at tool dyff:
```bash
dyff between --omit-header --set-exit-code a.yaml b.yaml 

(root level)
- one map entry removed:
  notfound: baz

values
  ⇆ order changed
    - foo: bar
  , bar: baz

    + bar: baz
  , foo: bar


same
  ± value change
    - notchanged
    + changed
```

We care about "same" where "value change" occurs. Both have key, but values are different.
We do not care about "order changed".
We care about "one map entry removed". (In one file but not the other).

Does that make snese?

> 
> Known differences to tpm2-tools: tpm2_readpublic has the following output for when scheme-halg is TPM2_ALG_ERROR:
> 
> ```yaml
> scheme-halg:
>   value: (null)
>   raw: 0x0
> ```
> 
> Which might be a bug

I think thats a bug. can you file a bug? How do you reproduce that?

whooo:
> > Open questions: > > * Should we encode it as YAML for the user or let them handle that part?
> 
> Yes return YAML. How else would you do it out of curiosity?
> 
As the other encoders/decoders, return dicts, strings, lists and integers so the user can use whatever serialization format they want (JSON, YAML etc) regardless of the actual structure
IIRC, someone at tpm.dev took the tools output and morphed into JSON for example, but I can always add two functions to_yaml and from_yaml which calls the encoder and then encodes/decodes it as YAML
> > * If we should do it, must it be character by character compatible? (See indentation in tpm2_getcap algorithms for example)
> 
> No, ordering/whitespace etc doesn't mater. Essentially if you load one document from tools and the other from python, as long as each key/value pair exists that's good enough, and nothing extra is hanging out.
> 
> Looking at tool dyff:
> 
> ....
> 
> We care about "same" where "value change" occurs. Both have key, but values are different. We do not care about "order changed". We care about "one map entry removed". (In one file but not the other).
> 
> Does that make snese?
> 
Yep, it does, I'm mostly worried about people using weird regex solutions in some shellscripts
> > Known differences to tpm2-tools: tpm2_readpublic has the following output for when scheme-halg is TPM2_ALG_ERROR:
> > ```yaml
> > scheme-halg:
> >   value: (null)
> >   raw: 0x0
> > ``` 
> >   
> > Which might be a bug
> 
> I think thats a bug. can you file a bug? How do you reproduce that?
I was thinking of doing a PR, I reproduce it by running tpm2_readpublic against a persistent key on my system, where the scheme is TPM2_ALG_NULL

If you can think of any tool I missed in the TODO list, please tell

williamcroberts:
<snip>
> > We care about "same" where "value change" occurs. Both have key, but values are different. We do not care about "order changed". We care about "one map entry removed". (In one file but not the other).
> > Does that make snese?
> 
> Yep, it does, I'm mostly worried about people using weird regex solutions in some shellscripts

They're banking on pretty fragile ordering. They should be using a proper YAML parser, so I wouldn't worry about them.

<snip>
> If you can think of any tool I missed in the TODO list, please tell

I added tpm2_readpublic and tpm2_print. @idesai do you know of any tools that are outputting a TPM data structure in YAML that's not in this list?

williamcroberts:
For item, "Block unsupported TPML types". Do we really want to do that or just say the base YAML encoder is what the tools should do if they spit out that data type to YAML? Should we block all types not being output in YAML? Is this encoder only for the YAML types or does it support the hex encoded digests as well?

whooo:
> I added tpm2_readpublic and tpm2_print. @idesai do you know of any tools that are outputting a TPM data structure in YAML that's not in this list?

tpm2_readpublic (without the names) and tpm2_print should be covered, as well as tpm2_getcap and tpm2_nvreadpublic

> For item, "Block unsupported TPML types". Do we really want to do that or just say the base YAML encoder is what the tools should do if they spit out that data type to YAML? Should we block all types not being output in YAML? Is this encoder only for the YAML types or does it support the hex encoded buffers as well?

For tools_encdec I think only formats that are in the tools already should be supported, otherwise they might become incompatible in the future

williamcroberts:
For testing we can run the tpm2-tools in the ci, (use the tcti option and set it to the tcti string in the test) and then grab the output and use the encoder and /decoder and semantically diff the difference. If tools not available, skip. Thoughts?

whooo:
> For testing we can run the tpm2-tools in the ci, (use the tcti option and set it to the tcti string in the test) and then grab the output and use the encoder and /decoder and semantically diff the difference. If tools not available, skip. Thoughts?

I think that would be good

whooo:
For the tools added to the list:
* tpm2_load, tpm2_loadexternal and tpm2_createak seem to only output the name(s), is that correct?
* tpm2_pcrread is hard to decode due to the two structures and the fact that a TPML_DIGEST can only contain eight digests
* tpm2_send I'm not sure what structure it outputs
* tpm2_getekcertificate doesn't seem to output any TPM structure, so I'll skip that one
* tpm2_quote, it's not possible to reconstruct a TPMT_SIGNATURE from the output, but I'll add a encoding method for it
* tpm2_print, it's not possible to reconstruct a TPMS_CONTEXT from the context yaml output and the output for a TPMS_QUOTE_INFO/TPMS_ATTEST is not valid YAML, so I'll probably remove both encoding and decoding support for those

idesai:
> > For the tools added to the list:
> > 
> > * tpm2_load, tpm2_loadexternal and tpm2_createak seem to only output the name(s), is that correct?
> 
> Yes I think that is correct.
> 
> > * tpm2_pcrread is hard to decode due to the two structures and the fact that a TPML_DIGEST can only contain eight digests
> 
> Well it does multiple reads, so I see two paths forward:
> 
> 1. decode could throw an error if too many selections in YAML
> 2. decode could return a list of TPML_DIGESTS.
> 
> I'm more in favor of option 2
> 
> > * tpm2_send I'm not sure what structure it outputs
> 
> binary in and out of formatted tpm command and response structures.
> 
> > * tpm2_getekcertificate doesn't seem to output any TPM structure, so I'll skip that one
> 
> Yep just the EK as a pem IIRC.
> 
> > * tpm2_quote, it's not possible to reconstruct a TPMT_SIGNATURE from the output, but I'll add a encoding method for it
> 
> Out's just some incomplete data?
> 
> > * tpm2_print, it's not possible to reconstruct a TPMS_CONTEXT from the context yaml output and the output for a TPMS_QUOTE_INFO/TPMS_ATTEST is not valid YAML, so I'll probably remove both encoding and decoding support for those
> 
> Hrm we probably need to fix those in tools @idesai were you aware of this issue?

Hmm, no actually. I will take a look. We may have to sync with keylime attestation folks prior to changes as they use these tools in their scripts.

whooo:
> > I need to figure out a good way for the API to take and output multiple structures in a sensible manner in both cases, suggestions are welcome
> 
> Both cases being encode and decode?
> 
> So to decode (YAML to type instance(s)) I would return a list. For encode (tpm2 type instance to YAML), I don't see how you would end up in that state. You'd just have a single type. I see the pcr read case where you could have N TPML_DIGESTS that you want to output as one YAML, you could have the decoder take variable args (*args) and then use that as additional inputs to concatenate with the YAML.
> 
The issue is that decode takes an instance (or type for integer types) to decode into, so now the API needs to handle taking multiple types at a time and understanding the context, but should be doable.
When it comes to encoding that output, both TPMS_PCR_SELECTION and TPML_DIGEST are needed to keep track of which digests belongs to which banks and currently the encoding API works after a type, just as decode.
I want to make sure that the API is usable and easy to understand by future users and having to keep track of which combinations of lists/tuples produce which type of output would make that harder.

It might be better to rewrite the encoder/decoder interface around which tool we want to decode/encode the output from, then we can require the right type of arguments for the expected input/output

> Would this patch get you what you need and if it's missing make hash alg an input to the decoder or throw error? Perhaps default it to sha256? For encoding, just output it like the patch does below.
> ... 
>
Yes, that would cover it, raising an exception would cause the least amount of surprises for when the hash alg is missing

> > So the issues I found while parsing TPMS_QUOTE_INFO and TPMS_ATTEST is that magic and type aren't parsed as hex but a string and a base 10 integer, same issue with firmwareVersion, also the pcrSelect is in some cases parsed as an octal, decimal or as hex integer depending on the value For TPMS_CONTEXT, it's missing the content of contextBlob
> 
> Interesting, I am not fully groking the complete details here but it sounds like a contextual mess. I would just skip them for now, especially the context blob, I don't see folks worried about encoding/decoding them, I see them loading them up and we have support for that.
> 
> I think it boils down to, if we cant encode/decode it with confidence, throw an unsupported/unimplemented exception for now.

I'll drop them both

williamcroberts:
@whooo thier is one task left, "Do visual inspection with actual YAML output from the encoder" are you ready there?

whooo:
> @whooo thier is one task left, "Do visual inspection with actual YAML output from the encoder" are you ready there?

Not entirely, and there is some new YAML output I need to add as well

whooo:
I'm not sure why the FAPI test fails, I haven't touched anything related to that test as far as I can see

whooo:
@williamcroberts, do you have any idea what the cause for the failing test might be? I'm unable to reproduce it locally with tpm2-tss from the master branch

williamcroberts:
> @williamcroberts, do you have any idea what the cause for the failing test might be? I'm unable to reproduce it locally with tpm2-tss from the master branch

No idea unfortunately, @JuergenReppSIT something changed on the master branch of FAPI that seems to be causing some issues.

JuergenReppSIT:
I also have no idea. The last commits were not related to to policy part.
Can you run this test with TSS2_LOG=fapi+trace to see whether Fapi_SetAuthCB is called to set the sign callback?
Or do you know what commit es responsible for this bug?

JuergenReppSIT:
In the trace the call of execute_policy_signed, which calls the callback function, did appear before the callback is set with Fapi_SetSignCB?

whooo:
> In the trace the call of execute_policy_signed, which calls the callback function, did appear before the callback is set with Fapi_SetSignCB?

It seems to do both, so there is a test for failure before setting the sign_callback at https://github.com/tpm2-software/tpm2-pytss/blob/707c23f91381b8f3f40cfacb1824314b9ecf6753/test/test_fapi.py#L750

whooo:
disabling FAPI support in the tools seem to have solved the issue, for an unknown reason

williamcroberts (approved):

Issue #330, opened 2022-02-26
Quote unmarshalling firmware value


iolivergithub:
Is the unmarshalling of a quote correct. The firmware value is unmarshalled as list of int and not as an int or str ?

```python3
# given suitable parameters
quote,signature = tpm.quote(
	  handle, pcrsels, extradata_to_use
	)
att,_ = TPMS_ATTEST.unmarshal( bytes(quote) )
enc = json_encdec() 
ae = enc.encode(att)
print("ae=",type(ae),"\n",ae)
```

returns

```
ae= <class 'dict'> 
 {'attested': {'pcrDigest': '38723a2e5e8a17aa7950dc008209944e898f69a7bd10a23c839d341e935fd5ca', 'pcrSelect': [{'hash': 'sha256', 'pcrSelect': [0, 1, 2, 3]}]}, 'clockInfo': {'clock': [3, 1014621385], 'resetCount': 75, 'restartCount': 0, 'safe': 1}, 'extraData': '49616e3132333435', 'firmwareVersion': [458837, 1166080], 'magic': 4283712327, 'qualifiedSigner': '000bf787a8fe2898c0c56f1c3a8629968282881c85e46e06365c097cc526e3425d9e', 'type': 'attest_quote'}
```

Tested with both Infineon SLB9670 and IBM SW TPM

Additionally  the type parameter with value "8018" (hex) is missing.   magic number is returned as an int and not the more familiar  "ff544347"   (minor issues!) - the latter can be convered as int( magic, 16 ) ...

PR #331, opened 2022-02-26
Encoding fixes

whooo:
Fixes https://github.com/tpm2-software/tpm2-pytss/issues/330


williamcroberts:
@whooo could you give me the notes to add to the changelog for this? I'm looking at merging these and cutting a release.

whooo:
> @whooo could you give me the notes to add to the changelog for this? I'm looking at merging these and cutting a release.

Perhaps something like "Ensure that TPM2_GENERATED.VALUE is encoded the same way as other constants" and "Add support to unmarshal simple TPM2B types (such as TPM2B_ATTEST and TPM2B_NAME) directly using the unmarshal method"

PR #332, opened 2022-03-01
ESAPI: add binding for Esys_TR_GetTpmHandle

williamcroberts:
The binding for Esys_TR_GetTpmHandle is missing, add it and a test.


tpm2_pytss/constants.py
```diff
@@ -316,6 +316,14 @@ def __str__(self):
...
 
 
+@TPM_FRIENDLY_INT._fix_const_type
+class TPM2_HANDLE(TPM_FRIENDLY_INT):
+    """TPM2_HANDLE is a raw handle type.
+    """
+
+    pass
+
+
```
whooo, line 326: TPM2_HANDLE is already defined in types.py

williamcroberts: Oh I was looking in constants.

whooo (approved):

Issue #335, opened 2022-04-14
1.1.0 fails to install with python 3,9 (Ubuntu)


iolivergithub:
Python 3.9, Ubuntu 21.10.   tpm2-tss installed using  apt install earlier ( >= 2.4  - at least U21.04 should have >=3.0.3 via apt install... ).   pip3 up to date, same error with pip.

First section shows via pip / pip3, second section installing from github  - here the error is more explicit:           raise RuntimeError("Require tss2-esapi to be installed and at least version 2.0.0") ... but  AFAIK tss2-esapi is 3.0.3 ?

~ $ sudo pip install tpm2-pytss==1.1.0
Collecting tpm2-pytss==1.1.0
  Using cached tpm2-pytss-1.1.0.tar.gz (163 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 /tmp/tmp13uy2rts_in_process.py prepare_metadata_for_build_wheel /tmp/tmpbxvd79x2
         cwd: /tmp/pip-install-9eq8nh4p/tpm2-pytss_e391132efa7e4adaa60f50ed694b513d
    Complete output (34 lines):
    Traceback (most recent call last):
      File "/tmp/tmp13uy2rts_in_process.py", line 280, in <module>
        main()
      File "/tmp/tmp13uy2rts_in_process.py", line 263, in main
        json_out['return_val'] = hook(**hook_input['kwargs'])
      File "/tmp/tmp13uy2rts_in_process.py", line 133, in prepare_metadata_for_build_wheel
        return hook(metadata_directory, config_settings)
      File "/tmp/pip-build-env-clxij1l9/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 188, in prepare_metadata_for_build_wheel
        self.run_setup()
      File "/tmp/pip-build-env-clxij1l9/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 174, in run_setup
        exec(compile(code, __file__, 'exec'), locals())
      File "setup.py", line 232, in <module>
        setup(
      File "/tmp/pip-build-env-clxij1l9/overlay/lib/python3.9/site-packages/setuptools/__init__.py", line 87, in setup
        return distutils.core.setup(**attrs)
[...]

PR #336, opened 2022-04-14
ci: test with and without fapi

williamcroberts:


williamcroberts:
relates to #335

whooo (approved):

PR #344, opened 2022-04-19
Python versions

whooo:
Drop 3.6 and add 3.10 with some fixes for distutils usage.
Also make sure the required version of cryptography is correct

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -7,12 +7,12 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10']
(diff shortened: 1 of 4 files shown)


williamcroberts:
@whooo could you submit another patch that "strigifies" all the variables in the matrix for both consistency and to ensure correctness as folks add and remove. This way YAML doesn't convert it to an int and drop things....

whooo:
> Looks like your 0 on the python 3.10 version is getting dropped (and you I see you force pushed as I was typing this lol)

Yeah, caught by YAML shenanigans once again

 

> @whooo could you submit another patch that "strigifies" all the variables in the matrix for both consistency and to ensure correctness as folks add and remove. This way YAML doesn't convert it to an int and drop things....

Done

williamcroberts (approved):

PR #365, opened 2022-06-29
Update README.md

noamraph:
The example said `pip install tpm2-ptss` instead of `pip install tpm2-pytss`, so it didn't work.


williamcroberts (approved):

williamcroberts (changes requested):
can you signoff your commit with `git commit --amend -s` or another method and re-push?

williamcroberts:
https://github.com/tpm2-software/tpm2-pytss/pull/378

Issue #357, opened 2022-05-18
1.1.0: sphinx warnings `reference target not found`


kloczek:
On building my packages I'm using `sphinx-build` command with `-n` switch which shows warmings about missing references. These are not critical issues.
```console
[tkloczko@devel-g2v tpm2-pytss-1.1.0]$ SETUPTOOLS_SCM_PRETEND_VERSION=1.1.0 PYTHONPATH=$PWD/build/$(cd build; ls -1d lib*) /usr/bin/sphinx-build -n -T -b man docs build/sphinx/man
Running Sphinx v4.5.0
Mocking tpm2_pytss._libtpm2_pytss
Warning: The Sphinx 'sphinx_rtd_theme' HTML theme was not found. Make sure you have the theme installed to produce pretty HTML output. Falling back to the default theme.
making output directory... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
myst v0.17.2: MdParserConfig(commonmark_only=False, gfm_only=False, enable_extensions=[], linkify_fuzzy_links=True, dmath_allow_labels=True, dmath_allow_space=True, dmath_allow_digits=True, dmath_double_inline=False, update_mathjax=True, mathjax_classes='tex2jax_process|mathjax_process|math|output_area', disable_syntax=[], all_links_external=False, url_schemes=('http', 'https', 'mailto', 'ftp'), ref_domains=None, highlight_code_blocks=True, number_code_blocks=[], title_to_header=False, heading_anchors=None, heading_slug_func=None, footnote_transition=True, sub_delimiters=('{', '}'), words_per_minute=200)
building [mo]: targets for 0 po files that are out of date
building [man]: all manpages
updating environment: [new config] 16 added, 0 changed, 0 removed
reading sources... [100%] utils
/home/tkloczko/rpmbuild/BUILD/tpm2-pytss-1.1.0/build/lib.linux-x86_64-cpython-38/tpm2_pytss/ESAPI.py:docstring of tpm2_pytss.ESAPI.ESAPI.load_blob:9: ERROR: Unknown target name: "type".
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
[...]

PR #369, opened 2022-09-17
Doc fixes

whooo:
Mostly some indentation fixes, removing "optional" from the Args type descriptions and make sure sphinx is aware of TSS2_Exception.

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/357

### DIFF
--- a/src/tpm2_pytss/ESAPI.py
+++ b/src/tpm2_pytss/ESAPI.py
@@ -81,9 +81,9 @@ class ESAPI:
         - TCP socket localhost:2321 (TPM simulator)
 
     Args:
-        tcti Union[TCTI, str]: The TCTI context used to connect to the TPM (may be None). This
-        is established using TCTILdr or a tpm2-tools style --tcti string in the format of
-        <tcti-name>:<tcti-conf> where :<tcti-conf> is optional. Defaults to None.
+        tcti (Union[TCTI, str]): The TCTI context used to connect to the TPM (may be None).
+          This is established using TCTILdr or a tpm2-tools style --tcti string in the format
+          of <tcti-name>:<tcti-conf> where :<tcti-conf> is optional. Defaults to None.
 
     Returns:
         An instance of the ESAPI class.
@@ -1687,7 +1687,7 @@ class ESAPI:
             TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
 
         Returns:
-            A Tuple[TPM2B_ECC_POINT, TPM2B_ECC_POINT] which is the zPoint Results of P := h[de]Qs
+            :A Tuple[TPM2B_ECC_POINT, TPM2B_ECC_POINT] which is the zPoint Results of P := h[de]Qs
             and pubPoint Generated ephemeral public point (Qe) respectively.
 
         C Function: Esys_ECDH_KeyGen
@@ -3118,7 +3118,7 @@ class ESAPI:
         available.
 
         Args:
-            curve_id (TPM2_ECC_CURVE): The curve for the computed ephemeral point .
+            curve_id (TPM2_ECC_CURVE): The curve for the computed ephemeral point.
             session1 (ESYS_TR): A session for securing the TPM command (optional). Defaults to ESYS_TR.NONE.
             session2 (ESYS_TR): A session for securing the TPM command (optional). Defaults to ESYS_TR.NONE.
             session3 (ESYS_TR): A session for securing the TPM command (optional). Defaults to ESYS_TR.NONE.
@@ -3129,7 +3129,7 @@ class ESAPI:
             TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
 
         Returns:
-            A Tuple[TPM2B_ECC_POINT, int] which is the Ephemeral public key Q := [r]G, known as Q,
+            :A Tuple[TPM2B_ECC_POINT, int] which is the Ephemeral public key Q := [r]G, known as Q,
             and the least-significant 16 bits of commitCount.
 
         C Function: Esys_EC_Ephemeral
@@ -6769,15 +6769,15 @@ class ESAPI:
     def load_blob(
         self, data: bytes, type_: int = _DEFAULT_LOAD_BLOB_SELECTOR
     ) -> ESYS_TR:
-        """load binary ESAPI object as binary blob. Supported are the types :const:`FAPI_ESYSBLOB.CONTEXTLOAD` and :const:`FAPI_ESYSBLOB.DESERIALIZE`.
+        """load binary ESAPI object as binary blob. Supported are the types :const:`constants.FAPI_ESYSBLOB.CONTEXTLOAD` and :const:`constants.FAPI_ESYSBLOB.DESERIALIZE`.
 
         Args:
             data (bytes): Binary blob of the ESAPI object to load.
-            type_ (int): :const:`FAPI_ESYSBLOB.CONTEXTLOAD` or :const:`FAPI_ESYSBLOB.DESERIALIZE`. Defaults to :const:`FAPI_ESYSBLOB.CONTEXTLOAD`
-                if FAPI is installed else :const: `FAPI_ESYSBLOB.DESERIALIZE`.
+            type_ (int): :const:`constants.FAPI_ESYSBLOB.CONTEXTLOAD` or :const:`constants.FAPI_ESYSBLOB.DESERIALIZE`. Defaults to :const:`constants.FAPI_ESYSBLOB.CONTEXTLOAD`
+              if FAPI is installed else :const:`constants.FAPI_ESYSBLOB.DESERIALIZE`.
 
         Raises:
-            ValueError: If type_ is not of an expected value.
+            ValueError: If type\_ is not of an expected value.
 
         Returns:
             ESYS_TR: The ESAPI handle to the loaded object.
--- a/src/tpm2_pytss/constants.py
+++ b/src/tpm2_pytss/constants.py
@@ -200,12 +200,33 @@ class TPM_FRIENDLY_INT(int):
         return self.__class__(int(self).__xor__(value))
 
     @staticmethod
-    def _fix_const_type(cls):
-        for k, v in vars(cls).items():
-            if not isinstance(v, int) or k.startswith("_"):
+    def _copy_and_set(dstcls, srccls):
+        """Copy class variables from srccls to dstcls
+
+        srccls must be a subclass on TPM_FRIENDLY_INT.
+        Sets the the variable type to dstcls when copying.
+        dstcls and srccls can be the same.
+        Skip setting destination if the dstcls variable already have the correct type.
+        """
+        if not issubclass(srccls, TPM_FRIENDLY_INT):
+            return
+        for k, v in vars(srccls).items():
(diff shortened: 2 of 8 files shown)


src/tpm2_pytss/ESAPI.py
```diff
@@ -1687,7 +1687,7 @@ def ecdh_key_gen(
             TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
 
         Returns:
-            A Tuple[TPM2B_ECC_POINT, TPM2B_ECC_POINT] which is the zPoint Results of P := h[de]Qs
+            :A Tuple[TPM2B_ECC_POINT, TPM2B_ECC_POINT] which is the zPoint Results of P := h[de]Qs
```
williamcroberts, line 1690: Why does this need a starting `:`?

src/tpm2_pytss/ESAPI.py
```diff
@@ -1770,7 +1770,7 @@ def ecdh_zgen(
 
     def ecc_parameters(
         self,
-        curve_id: TPM2_ECC_CURVE,
+        curve_id: TPM2_ECC,
```
williamcroberts, line 1773: Why change these?
```python
TPM2_ECC_CURVE = TPM2_ECC
```

Does it not generate the link correctly, should we do something like TPM2_ECC_CURVE extends TPM2_ECC?

src/tpm2_pytss/ESAPI.py
whooo: For some reason the documentation losses part of the line without, IIRC I tried a bunch of things but couldn't find anything better

src/tpm2_pytss/ESAPI.py
whooo: Yeah, the link doesn't work for some reason, I'm assuming it's because it's not a type but a variable according to the AST.
Do you mean make TPM2_ECC_CURVE a subclass of TPM2_ECC?

williamcroberts: Yep, and lets do that for all the "assigned types" to avoid this problem.
```python
class TPM2_ECC_CURVE(TPM2_ECC)
    pass
```

williamcroberts (changes requested):
Lets change all the "assigned types" to proper subclasses of the classes they were being assigned to.

williamcroberts (approved):

williamcroberts (changes requested):
Weird it blew something up
```
=========================== short test summary info ============================
FAILED test/test_crypto.py::CryptoTest::test_ecc_bad_curves - AssertionError: 'unsupported curve: 0' != 'unsupported curve: none'
- unsupported curve: 0
?                    ^
+ unsupported curve: none
?                    ^^^^
FAILED test/test_encoding.py::ToolsTest::test_tools_tpmt_public - AssertionError: {'nam[197 chars]e': '3', 'raw': 3}, 'kdfa-alg': {'value': None[384 chars]979'} != {'nam[197 chars]e': 'nist_p256', 'raw': 3}, 'kdfa-alg': {'valu[392 chars]979'}
  {'attributes': {'raw': 262258,
                  'value': 'fixedtpm|fixedparent|sensitivedataorigin|userwithauth|sign'},
-  'curve-id': {'raw': 3, 'value': '3'},
?                                   ^

+  'curve-id': {'raw': 3, 'value': 'nist_p256'},
?                                   ^^^^^^^^^
```

williamcroberts:
> Yeah, I assuming it's because the inherited class variables aren't viewed as part of the class when we introspect them in all the different class methods

Yeah that's what it looks like offhand... I guess do we want to just take your approach or update the logic in the handler to go up a class and a try?

whooo:
> > Yeah, I assuming it's because the inherited class variables aren't viewed as part of the class when we introspect them in all the different class methods
> 
> Yeah that's what it looks like offhand... I guess do we want to just take your approach or update the logic in the handler to go up a class and a try?

One option is to simply copy all the class variables to the new class, but that would of course mean two very copies of the same class more or less, I'll see if I can wrangle sphinx in any way or reuse the constant decorator

whooo:
I did an ugly hack in `_fix_const_type`, so if the parent class isn't one of the base friendly int classes it will copy the variables from the parent to the child.
I'll take some time to see if I can find a better solution.

williamcroberts:
> I did an ugly hack in `_fix_const_type`, so if the parent class isn't one of the base friendly int classes it will copy the variables from the parent to the child. I'll take some time to see if I can find a better solution.

I wonder if you can look at the full subclass list:
  - https://stackoverflow.com/questions/31028237/getting-all-superclasses-in-python-3

whooo:
That was a good find, think I got it right now with using  `__mro__`.
Also found TPM2_ALG_ID and did the same fix as for TPM2_ECC_CURVE

williamcroberts:
> That was a good find, think I got it right now with using `__mro__`. Also found TPM2_ALG_ID and did the same fix as for TPM2_ECC_CURVE

Don't forget TPM2_CLOCK_ADJUST

williamcroberts:
> > That was a good find, think I got it right now with using `__mro__`. Also found TPM2_ALG_ID and did the same fix as for TPM2_ECC_CURVE
> 
> Don't forget TPM2_CLOCK_ADJUST

This is what I used to find all of them, hopefully:
```bash
grep -nE '^[A-Z0-9_]+\s=\s[A-Z0-9_]+' *.py
```

src/tpm2_pytss/constants.py
```diff
@@ -200,12 +200,28 @@ def __xor__(self, value):
...
+            fv = dstcls(v)
+            setattr(dstcls, k, fv)
+
+    @staticmethod
+    def _fix_const_type(cls):
+        TPM_FRIENDLY_INT._copy_and_set(cls, cls)
+        # check for superclasses which are subclasses of TPM_FRIENDLY_INT, but isn't
+        # a TPM_FRIENDLY_INT or TPMA_FRIENDLY_INTLIST and copy the constants from the
+        # parent class. This enables good inheritance for the different constants types.
+        for sc in cls.__mro__[1:]:
```
williamcroberts, line 221: doesn't this work? We essentially don't care about object, int, and the superclass helper which are the top 3 in the inheritance list (see output of mro slice below). I would like to avoid hardcoding class names, but I guess relying on inheritance structure could be pretty brittle too. I guess its two versions of the same problem. Your thoughts?

```python
for sc in cls.__mro__[1:-3]:
```

```python3
>>> TPM2_ECC.__mro__[1:-3]
()
>>> TPM2_ALG_ID.__mro__[1:-3]
(<class 'tpm2_pytss.constants.TPM2_ALG'>,)
```

williamcroberts: @whooo this patch ontop of yours works for me:
```git
diff --git a/src/tpm2_pytss/constants.py b/src/tpm2_pytss/constants.py
index a2acd127e321..520f22b52b12 100644
--- a/src/tpm2_pytss/constants.py
+++ b/src/tpm2_pytss/constants.py
@@ -214,13 +214,18 @@ class TPM_FRIENDLY_INT(int):
 
     @staticmethod
     def _fix_const_type(cls):
-        TPM_FRIENDLY_INT._copy_and_set(cls, cls)
-        # check for superclasses which are subclasses of TPM_FRIENDLY_INT, but isn't
-        # a TPM_FRIENDLY_INT or TPMA_FRIENDLY_INTLIST and copy the constants from the
-        # parent class. This enables good inheritance for the different constants types.
-        for sc in cls.__mro__[1:]:
-            if sc in (TPM_FRIENDLY_INT, TPMA_FRIENDLY_INTLIST):
-                break
+        # We want to copy class variables from the base class and any "intermediate" classes.
+        #
+        # Skip the:
+        # - the root "object" class
+        # - Helper classes TPM_FRIENDLY_INT or TPM_FRIENDLY_INTLIST
+        # consider the following examples showing inheritance with
+        # a helper class from the list above and without.
+        # >>> TPM2_ECC.__mro__[:-3]
+        # (<class 'tpm2_pytss.constants.TPM2_ECC'>,)
+        # >>> TPM2_ALG_ID.__mro__[:-3]
+        # (<class 'tpm2_pytss.constants.TPM2_ALG_ID'>, <class 'tpm2_pytss.constants.TPM2_ALG'>)
+        for sc in cls.__mro__[:-3]:
             TPM_FRIENDLY_INT._copy_and_set(cls, sc)
         return cls
 ```

Comment (whooo) [reply src/tpm2_pytss/constants.py:221]: My worry is that it would create surprises in the future if it doesn't follow a strict inheritance scheme, lets say someone wants to add a mixin class and writes:
```python3
@TPM_FRIENDLY_INT._fix_const_type
TPM2_COOL_CONSTANT(MyMixin, TPM2_ECC):
    pass
```
That would break that flow and it wouldn't be obvious why without digging a bit in the code.
Looking at my current code it would still copy some things from the mixin, which it shouldn't.
So what about, find the first superclass that is a subclass of TPM_FRIENDLY_INT but isn't TPM_FRIENDLY_INT or TPMA_FRIENDLY_INTLIST and copy the constants from that?

I'm also looking at making sure that one could override a value in a subclass

williamcroberts: > My worry is that it would create surprises in the future if it doesn't follow a strict inheritance scheme, lets say someone wants to add a mixin class and writes:
> 
> ```python
> @TPM_FRIENDLY_INT._fix_const_type
> TPM2_COOL_CONSTANT(MyMixin, TPM2_ECC):
>     pass
> ```
> 
> That would break that flow and it wouldn't be obvious why without digging a bit in the code. Looking at my current code it would still copy some things from the mixin, which it shouldn't. So what about, find the first superclass that is a subclass of TPM_FRIENDLY_INT but isn't TPM_FRIENDLY_INT or TPMA_FRIENDLY_INTLIST and copy the constants from that?
> 
> I'm also looking at making sure that one could override a value in a subclass

If fine with keeping it dirt simple as either case we assume a certain inheritance hierarchy and your code is more verbose about that.

```diff
diff --git a/src/tpm2_pytss/constants.py b/src/tpm2_pytss/constants.py
index a2acd127e321..2ea183d271eb 100644
--- a/src/tpm2_pytss/constants.py
+++ b/src/tpm2_pytss/constants.py
@@ -214,11 +214,10 @@ class TPM_FRIENDLY_INT(int):
 
     @staticmethod
     def _fix_const_type(cls):
-        TPM_FRIENDLY_INT._copy_and_set(cls, cls)
         # check for superclasses which are subclasses of TPM_FRIENDLY_INT, but isn't
         # a TPM_FRIENDLY_INT or TPMA_FRIENDLY_INTLIST and copy the constants from the
         # parent class. This enables good inheritance for the different constants types.
-        for sc in cls.__mro__[1:]:
+        for sc in cls.__mro__:
             if sc in (TPM_FRIENDLY_INT, TPMA_FRIENDLY_INTLIST):
                 break
             TPM_FRIENDLY_INT._copy_and_set(cls, sc)
```
What do you think of this simplification, we copy the baseclass as well as the parent classes so long as they are not TPM_FRIENDLY...?

whooo: Looks better to me, I'll just split out the changes to how `_fix_const_type` works into its own commit

williamcroberts (approved):

PR #385, opened 2022-10-28
ci: cache dependencies

whooo:
Signed-off-by: Erik Larsson <who+github@cnackers.org>


williamcroberts:
Ahh actions cache, good idea.

PR #387, opened 2022-10-31
Cache deps

whooo:
Cache dependencies between runs, excluding jobs that use the tpm2-tss master branch there is a reduction in runtime from 5-7 minutes to 3-4 minutes, this in combination with https://github.com/tpm2-software/tpm2-pytss/pull/388 should decrease time to feedback for pull requests.

The only issue is that the cache is only valid for one week, so a new pull request after a week of inactivity will still take some time.

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -5,6 +5,12 @@ if [ -d "${HOME}/.local/bin" ]; then
   export PATH="${HOME}/.local/bin:${PATH}"
 fi
 
+# Setup environment for using cached libraries/binaries
+export CI_DEPS_PATH="${HOME}/cideps"
+export PATH="${PATH}:${CI_DEPS_PATH}/bin"
+export PKG_CONFIG_PATH="${CI_DEPS_PATH}/lib/pkgconfig"
+export LD_LIBRARY_PATH="${CI_DEPS_PATH}/lib/"
+
 SRC_ROOT=${SRC_ROOT:-"${PWD}"}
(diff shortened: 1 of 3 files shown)


whooo:
@williamcroberts this PR is ready now, but I assume this should go in after 1.3.0 as with https://github.com/tpm2-software/tpm2-pytss/pull/388

.ci/run.sh
```diff
@@ -5,6 +5,12 @@ if [ -d "${HOME}/.local/bin" ]; then
   export PATH="${HOME}/.local/bin:${PATH}"
 fi
 
+# Setup environment for using cached libraries/binaries
+export CI_DEPS_PATH="${HOME}/cideps"
+export PATH="${PATH}:${CI_DEPS_PATH}/bin"
+export PKG_CONFIG_PATH="${CI_DEPS_PATH}/lib/pkgconfig"
+export LD_LIBRARY_PATH="${CI_DEPS_PATH}/lib/"
```
williamcroberts, line 12: I would append to `PKG_CONFIG_PATH` and `LD_LIBRARY_PATH` over nuking whats there. This way if someone runs `run.sh` locally, they can set stuff to however they need it.

williamcroberts (changes requested):

.ci/run.sh
whooo: Fixed

whooo:
Will open a new PR

PR #393, opened 2022-11-08
ESAPI.py: Adding 1.59 updates

malikabhi05:
Signed-off-by: malikabhi05 <abhishek.malik@intel.com>

### DIFF
--- a/src/tpm2_pytss/ESAPI.py
+++ b/src/tpm2_pytss/ESAPI.py
@@ -2179,6 +2179,68 @@ class ESAPI:
         )
         return TPM2B_DIGEST(_get_dptr(out_hMAC, lib.Esys_Free))
 
+    def mac(
+        self,
+        handle: ESYS_TR,
+        buffer: Union[TPM2B_MAX_BUFFER, bytes, str],
+        hash_alg: TPM2_ALG,
+        session1: ESYS_TR = ESYS_TR.PASSWORD,
+        session2: ESYS_TR = ESYS_TR.NONE,
+        session3: ESYS_TR = ESYS_TR.NONE,
+    ) -> TPM2B_DIGEST:
+        """Invoke the TPM2_MAC command.
+
+        This function invokes the TPM2_MAC command in a one-call
+        variant. This means the function will block until the TPM response is
+        available.
+
+        Args:
+            handle (ESYS_TR): Handle for the symmetric signing key providing the HMAC key.
+            buffer (Union[TPM2B_MAX_BUFFER, bytes, str]): MAC data.
+            hash_alg (TPM2_ALG): Algorithm to use for MAC.
+            session1 (ESYS_TR): A session for securing the TPM command (optional). Defaults to ESYS_TR.PASSWORD.
+            session2 (ESYS_TR): A session for securing the TPM command (optional). Defaults to ESYS_TR.NONE.
+            session3 (ESYS_TR): A session for securing the TPM command (optional). Defaults to ESYS_TR.NONE.
+
+        Raises:
+            TypeError: If a parameter is not of an expected type.
+            ValueError: If a parameter is not of an expected value.
+            TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
+
+        Returns:
+            A TPM2B_DIGEST result of the MAC.
+
+        C Function: Esys_MAC
+
+        TPM Command: TPM2_MAC
+        """
+        if not _lib_version_atleast("tss2-esys", "3.2.0-167-gc17e3989"):
+            raise NotImplementedError("MAC api not supported below ESAPI v4")
+
+        _check_handle_type(handle, "handle")
+        _check_handle_type(session1, "session1")
+        _check_handle_type(session2, "session2")
+        _check_handle_type(session3, "session3")
+
+        _check_friendly_int(hash_alg, "hash_alg", TPM2_ALG)
+
+        buffer_cdata = _get_cdata(buffer, TPM2B_MAX_BUFFER, "buffer")
+
+        out_MAC = ffi.new("TPM2B_DIGEST **")
+        _chkrc(
+            lib.Esys_MAC(
+                self._ctx,
+                handle,
+                session1,
+                session2,
+                session3,
+                buffer_cdata,
+                hash_alg,
+                out_MAC,
+            )
+        )
+        return TPM2B_DIGEST(_get_dptr(out_MAC, lib.Esys_Free))
+
     def get_random(
         self,
         bytes_requested: int,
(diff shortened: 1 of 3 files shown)


src/tpm2_pytss/ESAPI.py
```diff
@@ -2179,6 +2179,65 @@ def hmac(
...
+            ValueError: If a parameter is not of an expected value.
+            TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
+
+        Returns:
+            A TPM2B_DIGEST result of the MAC.
+
+        C Function: Esys_MAC
+
+        TPM Command: TPM2_MAC
+        """
```
williamcroberts, line 2216: I think we need to ensure that `lib.Esys_MAC` exists, because it will fail on older versions that don't have this.

Probably something like:
```python
if not hasattr(lib, "Esys_MAC"):
    raise NotImplementedError("ESAPI Does not have MAC Interface")
```

@whooo do you have a preferred way of doing this?

src/tpm2_pytss/ESAPI.py
```diff
@@ -2331,6 +2390,71 @@ def hmac_start(
...
+            TypeError: If a parameter is not of an expected type.
+            ValueError: If a parameter is not of an expected value.
+            TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
+
+        Returns:
+            An ESYS_TR handle of ESYS resource for TPMI_DH_OBJECT.
+
+        C Function: Esys_MAC_Start
+
+        TPM Command: TPM2_MAC_Start
```
williamcroberts, line 2426: Same for this command, make sure the attribute exists.

williamcroberts (changes requested):

src/tpm2_pytss/ESAPI.py
whooo: Perhaps doing something similar to https://github.com/tpm2-software/tpm2-pytss/blob/master/src/tpm2_pytss/constants.py#L870 would be good, it would make to code look a bit weird but would make checking for support easier

williamcroberts: Yeah I like that better, using `if _lib_version_atleast("tss2-esapi", "3.0.0"):`

williamcroberts: Grr, this is where it gets hard, Since we're doing this ahead of tag 4.0, we need to check if attribute exists. We can switch this later....

src/tpm2_pytss/ESAPI.py
```diff
@@ -2179,6 +2179,68 @@ def hmac(
...
+        _check_handle_type(handle, "handle")
+        _check_handle_type(session1, "session1")
+        _check_handle_type(session2, "session2")
+        _check_handle_type(session3, "session3")
+
+        _check_friendly_int(hash_alg, "hash_alg", TPM2_ALG)
+
+        buffer_cdata = _get_cdata(buffer, TPM2B_MAX_BUFFER, "buffer")
+
+        out_hMAC = ffi.new("TPM2B_DIGEST **")
```
williamcroberts, line 2229: out_HMAC, I think this needs to be out_MAC to work. My question is how are the tests passing :-p

williamcroberts (changes requested):

test/test_esapi.py
```diff
@@ -1435,6 +1436,63 @@ def test_hmac(self):
         with self.assertRaises(TypeError):
             self.ectx.hmac(primaryHandle, inData, TPM2_ALG.SHA256, session3=45.6)
 
+    def test_mac(self):
```
williamcroberts, line 1439: Oh it works because your raising an Exception of NotImplementedError?

Perhaps this should decorate this function as:
```python
    @pytest.mark.skipif(
        not _lib_version_atleast("tss2-esapi", "4.0.0"), reason="Requires tss2-esapi 4.0 or greater"
    )
```

src/tpm2_pytss/ESAPI.py
malikabhi05: I think thats because the first test is to check for v >= 4.0 and we dont have a 4.0 job so it never goes past that. I'll fix the out_HMAC

williamcroberts (changes requested):
I guess we need to figure out how to get this to test now against master without checking for 4.0.
I think for not use hasattr to determine if we do it, and file a ticket to modify them. Perhaps add a TODO: VERSCHECK or some unique string in the code so we know where to fix them all.

whooo:
> I guess we need to figure out how to get this to test now against master without checking for 4.0. I think for not use hasattr to determine if we do it, and file a ticket to modify them. Perhaps add a TODO: VERSCHECK or some unique string in the code so we know where to fix them all.

Using a recent git version might be enough, that is what I did for the policy parts and I'm already thinking of doing a PR switching all git versions to 4.0.0 when it's released

williamcroberts:
> > I guess we need to figure out how to get this to test now against master without checking for 4.0. I think for not use hasattr to determine if we do it, and file a ticket to modify them. Perhaps add a TODO: VERSCHECK or some unique string in the code so we know where to fix them all.
> 
> Using a recent git version might be enough, that is what I did for the policy parts and I'm already thinking of doing a PR switching all git versions to 4.0.0 when it's released

Ahh I didn't know that would work. @malikabhi05 look at how @whooo did it, lets do it like that:
```python
if not _lib_version_atleast("tss2-policy", "3.2.0-63-gdcdc8412"):
    raise NotImplementedError(
        "tss2-policy not installed or version is less then 3.2.0-63-gdcdc8412"
    )
```
So you can just use some recent `git describe --always` output from tpm2-tss. I just merged your final 1.59 PR in tpm2-tss, so lets use that as the "minimum":
```bash
git describe --always
3.2.0-167-gc17e3989bf2f
```

src/tpm2_pytss/ESAPI.py
```diff
@@ -2179,6 +2179,66 @@ def hmac(
...
+
+        Returns:
+            A TPM2B_DIGEST result of the MAC.
+
+        C Function: Esys_MAC
+
+        TPM Command: TPM2_MAC
+        """
+
+        _check_handle_type(handle, "handle")
```
williamcroberts, line 2218: @malikabhi05 here do the libversion check and raise NotImplementedError if it's too old.

src/tpm2_pytss/ESAPI.py
williamcroberts: @malikabhi05 here do the libversion check and raise NotImplementedError if it's too old.

test/test_esapi.py
```diff
@@ -1435,6 +1437,67 @@ def test_hmac(self):
...
+
+        inSensitive = TPM2B_SENSITIVE_CREATE(TPMS_SENSITIVE_CREATE())
+
+        primaryHandle = self.ectx.create_primary(inSensitive, inPublic)[0]
+
+        # Test for tss v < 4.0.0
+        if not _lib_version_atleast("tss2-esapi", "4.0.0"):
+            with self.assertRaises(NotImplementedError):
+                self.ectx.mac(primaryHandle, b"1234", TPM2_ALG.SHA256)
+            return
```
williamcroberts, line 1462: Is this test block ever running, the NotImplementedError is never thrown by ESAPI python code.

williamcroberts (changes requested):

src/tpm2_pytss/ESAPI.py
```diff
@@ -2179,6 +2179,68 @@ def hmac(
...
+            TSS2_Exception: Any of the various TSS2_RC's the lower layers can return.
+
+        Returns:
+            A TPM2B_DIGEST result of the MAC.
+
+        C Function: Esys_MAC
+
+        TPM Command: TPM2_MAC
+        """
+        if not _lib_version_atleast("tss2-esys", "4.0.0"):
```
williamcroberts, line 2217: Version should be 3.2.0-167-gc17e3989bf2f right?

src/tpm2_pytss/ESAPI.py
```diff
@@ -2331,6 +2393,74 @@ def hmac_start(
...
+
+        Returns:
+            An ESYS_TR handle of ESYS resource for TPMI_DH_OBJECT.
+
+        C Function: Esys_MAC_Start
+
+        TPM Command: TPM2_MAC_Start
+        """
+
+        if not _lib_version_atleast("tss2-esys", "4.0.0"):
```
williamcroberts, line 2432: version should be: 3.2.0-167-gc17e3989bf2f

src/tpm2_pytss/ESAPI.py
```diff
@@ -5960,6 +6090,206 @@ def get_capability(
...
+
+        Returns:
+            A Tuple[bool, TPML_AC_CAPABILITIES] which is the Flag to indicate if there are more values of this type
+            and the capability data respectively.
+
+        C Function: Esys_AC_GetCapability
+
+        TPM Command: TPM2_AC_GetCapability
+        """
+        if not _lib_version_atleast("tss2-esys", "4.0.0"):
```
williamcroberts, line 6128: version should be: 3.2.0-167-gc17e3989bf2f

src/tpm2_pytss/ESAPI.py
```diff
@@ -5960,6 +6090,206 @@ def get_capability(
...
+
+        Returns:
+            A TPMS_AC_OUTPUT result, may include AC specific data or information about an error.
+
+        C Function: Esys_AC_Send
+
+        TPM Command: TPM2_AC_Send
+        """
+
+        if not _lib_version_atleast("tss2-esys", "4.0.0"):
```
williamcroberts, line 6199: Same version should be 3.2.0-167-gc17e3989bf2f

williamcroberts (changes requested):
All the 4.0.0 version checks... i didn't comment on all of them, should be 3.2.0-167-gc17e3989bf2f

williamcroberts (approved):

PR #396, opened 2022-11-09
Pure python TCTI Interface

williamcroberts:
Allow for TCTIs to be written in Pure python. This adds much needed Documentation around the TCTI interface and sanity on some types.

I'd like to pull this and bump the release version to 2.0 as we will have some minor API tweaks.

Note: This breaks some of the TCTI API, which AFAIK was unused. The
changes are:
- magic: from int to a bytes, which behaves a little more nicely IMHO.
- get_poll_handles: from a tuple of CData objects into a tuple of
  PollData objects.

### DIFF
--- a/test/test_tcti.py
+++ b/test/test_tcti.py
@@ -7,10 +7,41 @@ from tpm2_pytss import *
 from .TSS2_BaseTest import TSS2_EsapiTest
 
 
+class MyTCTI(PyTCTI):
+    def __init__(self, subtcti, magic=None):
+        self._tcti = subtcti
+        if magic is not None:
+            super().__init__(magic=magic)
+        else:
+            super().__init__()
+
+    def do_transmit(self, command):
+        self._tcti.transmit(command)
+
+    def do_receive(self, timeout):
+        return self._tcti.receive()
+
+    def do_cancel(self):
+        self._tcti.cancel()
+
+    def do_get_poll_handles(self):
+        return self._tcti.get_poll_handles()
+
+    def do_set_locality(self, locality):
+        self._tcti.set_locality(locality)
+
+    def do_make_sticky(self, handle, is_sticky):
+        if self._tcti is not None:
+            self._tcti.make_sticky(handle, is_sticky)
+
+        if self._error is not None:
+            raise self._error
+
+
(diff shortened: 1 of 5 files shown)


williamcroberts:
@AndreasFuchsTPM you wanted a server/client rest API, here's the Python backbone to write something in haha....

test/test_tcti.py
```diff
@@ -73,6 +104,103 @@ def test_tctildr(self):
...
+
+        t.transmit(b"\x80\x01\x00\x00\x00\x0C\x00\x00\x01\x44\x00\x00")
+        t.cancel()
+
+    def test_custom_pytcti_finalize(self):
+        # This is NOP for the PyTCTI but nothing should break
+        # TODO maybe it should implement it?
+        t = MyTCTI(self.tcti)
+        t.finalize()
+
```
whooo, line 138: Would be good to have finalize implemented, it's a clear signal from the caller that the instance is not needed anymore

whooo (approved):

test/test_tcti.py
williamcroberts: > Would be good to have finalize implemented, it's a clear signal from the caller that the instance is not needed anymore

My thought that `__del__` on the Python object was the clear signal, no?

I also added a test for the context manager and just mocked finalize to ensure that it is called. It just NOPs in the Python TCTI Implementation as the Finalize Function Pointer is NULL.

My inclination is to not implement it, but it's always easy to add if someone bumps into it... @whooo play devils advocate here, I could be convinced to add it now with a good argument.

williamcroberts: I'll take any changes as a next patchset in the series... merged now, but lets keep the discussion going.

williamcroberts: > The issue with `__del__` is that you don't know when it's actually called, as that is up to the garbage collector, so finalize would be more like `__exit__` or ESAPI.close. There is also the case where you use a PyTCTI subclass as a proxy TCTI, where you might want to proxy finalize. An implementation could of course ignore it by using a default method from PyTCTI.

Ahh yeah consider you have a TCTI that blocks other users from the resource, you want an explicitly timed event so others can establish a TCTI connection. Sold.

Issue #463, opened 2022-11-19
Convert TPMT_SIGNATURE to DER


JuergenReppSIT:
I tried do verify a signature, returned by  `esys.get_time` , with `fapi.verify_signature`.
`fapi.verify_signature` expects a DER signature. The tpm2 tool command `tpm2_gettime` has
the option `--format=plain` for this purpose.
Is there a way to convert a TPMT_SIGNATURE to DER with tpm2_pytss?
`

whooo:
There is no public method for that, you can have a look at https://github.com/tpm2-software/tpm2-pytss/blob/1465ae2f2bf2753b4d9dccd71d66f2088cde263d/src/tpm2_pytss/internal/crypto.py#L430 to see how we handle signature verification using the cryptography module for ECDSA signatures.

Adding the `__bytes__` method to TPMT_SIGNATURE which would return the underlying bytes for RSA and HMAC signatures and the DER encoding for ECSA signatures could be useful, would that work for `fapi.verify_signaure`?

JuergenReppSIT:
> Adding the `__bytes__` method to TPMT_SIGNATURE which would return the underlying bytes for RSA and HMAC signatures and the DER encoding for ECSA signatures could be useful, would that work for `fapi.verify_signaure`?

Yes, that would be great. Currently I did the verification using tss2_verifysignature with the  DER output from tpm2_gettime and it worked. With tpm2_pytss the implementation seemed to be easier because of good features to access data in the binary blobs.

PR #464, opened 2022-11-19
types: add __bytes__ to TPMT_SIGNATURE

whooo:
Make calling bytes(sig) return the underlying signature bytes. Useful to pass the signature to other APIs/libraries who don't use TPM structures.

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/463

### DIFF
--- a/src/tpm2_pytss/internal/crypto.py
+++ b/src/tpm2_pytss/internal/crypto.py
@@ -402,6 +402,21 @@ def _get_digest_size(alg):
     return dt.digest_size
 
 
+def _get_signature_bytes(sig):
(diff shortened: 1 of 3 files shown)


whooo:
@JuergenReppSIT, could you test with this?

JuergenReppSIT:
> @JuergenReppSIT, could you test with this?

@whooo 
`der_sig=signature.__bytes__()`
`fapi.verify_signature("/ext/time_sign", digest, der_sig)`

works :+1:  thank you so much for solving it so quickly!

williamcroberts (dismissed):

williamcroberts (approved):

Issue #466, opened 2022-11-20
FAPI Documentation


JuergenReppSIT:
The documentation of the FAPI API can be produced with `./scripts/docs.sh `
On https://tpm2-pytss.readthedocs.io/en/latest/fapi.html there is no documentation of the FAPI functions.
Is there a reason for that?

whooo:
I wonder if it's due to library version checks we have for FAPI (and other parts), so for local doc generation sphinx can import the module but for automatic doc generation FAPI won't be imported.

PR #467, opened 2022-11-20
docs: use fake library versions for readthedocs

whooo:
As readthedocs documentation generation uses the source tree and not a build, set fake versions to make sure that he version dependent parts are included in the public docs.

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/466

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,6 +22,15 @@ from sphinx.util import logging
 
 logger = logging.getLogger(__name__)
 
+
+class FakeVersions:
+    _versions = {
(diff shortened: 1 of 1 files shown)


whooo:
@williamcroberts, I don't have much experience with readthedocs, does it look sane to you?
I tried this on a fresh checkout of current master, with this applied running `READTHEDOCS=1 ./scripts/docs.sh` works without tpm2-pytss being installed

williamcroberts (approved):

williamcroberts (approved):

Issue #465, opened 2022-11-19
tpm2-pytss install suddenly not working (Ubuntu 20.04, WSL2)


Jarkre:
I tried to install tpm2_pytss to Ubuntu 20.04 but the following error message is shown:

    cffi.CDefError: cannot parse "#if !defined(DISABLE_VENDOR)"
    <cdef source string>:2193:2: before: if

What could be the reason for this?

The full log:

pip install tpm2_pytss --no-cache-dir

Collecting tpm2_pytss
  Downloading tpm2-pytss-1.2.0.tar.gz (178 kB)
     |████████████████████████████████| 178 kB 5.5 MB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 /tmp/tmpxo948zq3 prepare_metadata_for_build_wheel /tmp/tmpgjlmslw0
         cwd: /tmp/pip-install-nqwdlqvq/tpm2-pytss
    Complete output (66 lines):
    adding path: /tmp/pip-install-nqwdlqvq/tpm2-pytss/scripts
    Traceback (most recent call last):
      File "/usr/local/lib/python3.8/dist-packages/cffi/cparser.py", line 336, in _parse
        ast = _get_parser().parse(fullcsource)
      File "/usr/local/lib/python3.8/dist-packages/pycparser/c_parser.py", line 147, in parse
        return self.cparser.parse(
      File "/usr/local/lib/python3.8/dist-packages/pycparser/ply/yacc.py", line 331, in parse
        return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
      File "/usr/local/lib/python3.8/dist-packages/pycparser/ply/yacc.py", line 1199, in parseopt_notrack
        tok = call_errorfunc(self.errorfunc, errtoken, self)
      File "/usr/local/lib/python3.8/dist-packages/pycparser/ply/yacc.py", line 193, in call_errorfunc
        r = errorfunc(token)
      File "/usr/local/lib/python3.8/dist-packages/pycparser/c_parser.py", line 1931, in p_error
        self._parse_error(
      File "/usr/local/lib/python3.8/dist-packages/pycparser/plyparser.py", line 67, in _parse_error
        raise ParseError("%s: %s" % (coord, msg))
[...]

PR #468, opened 2022-11-20
scripts: remove "#if ..." when preparing headers

whooo:
A recent commit to tpm2-tss adds a "#if ..." to a header, which we where not removing, causing problems for CFFI, so remove any "#if ...".

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/465

### DIFF
--- a/scripts/prepare_headers.py
+++ b/scripts/prepare_headers.py
@@ -21,6 +21,7 @@ def remove_common_guards(s):
 
     # Remove includes and guards
     s = re.sub("#ifndef.*", "", s)
+    s = re.sub("#if .*", "", s)
     s = re.sub("#define .*_H_*?\n", "", s)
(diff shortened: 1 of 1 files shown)


williamcroberts:
@whooo  this worked for me, which is a bit suspect since we're cleaning out all `#if`'s here.
  - https://github.com/tpm2-software/tpm2-tss/pull/2498

I guess we just pull this instead.

williamcroberts:
tpm2-tools is fixed by PR: https://github.com/tpm2-software/tpm2-tools/pull/3167

williamcroberts:
I merged this as https://github.com/tpm2-software/tpm2-pytss/pull/470/commits/077bc9eb691558b135defd87fe98dc9de4b31977

Issue #480, opened 2022-12-01
{AttributeError}'TPMT_PUBLIC' object has no attribute '_cdata'


jwz104:
I have an issue when I convert a public key to a `TPM2B_PUBLIC`. The following code fails (sometimes):
```python
# tpm_cert is a cryptography certificate instance of an ECC EK certificate.
endorsement_key = TPM2B_PUBLIC.from_pem(
    tpm_cert.public_bytes(encoding=serialization.Encoding.DER),
    symmetric="aes128cfb", 
)
```

I get the following exception:
```
  File "/usr/local/lib/python3.9/site-packages/tpm2_pytss/types.py", line 1195, in from_pem
    pa = TPMT_PUBLIC.from_pem(
  File "/usr/local/lib/python3.9/site-packages/tpm2_pytss/types.py", line 976, in from_pem
    _public_from_encoding(data, p, password=password)
  File "/usr/local/lib/python3.9/site-packages/tpm2_pytss/internal/crypto.py", line 154, in _public_from_encoding
    obj.type = TPM2_ALG.ECC
  File "/usr/local/lib/python3.9/site-packages/tpm2_pytss/types.py", line 140, in __setattr__
    _cdata = object.__getattribute__(self, "_cdata")
AttributeError: 'TPMT_PUBLIC' object has no attribute '_cdata'
```

The exception occurs when I call any field (e.g. `type`) on the `TPMT_PUBLIC` instance. See the debugger below:
![afbeelding](https://user-images.githubusercontent.com/6162114/205036180-500f88af-396b-4072-abc8-de731ffe70b1.png)

My setup might be a bit weird:
- I run on Mac OS 12.6
- I using a remote Python interpreter from a Docker container so I have TPM2-TSS.
- I'm using Python 3.9
- Using version 2.4.0 of tpm2-pytss

The weird thing is that I sometimes do not get the exception. Does anyone know how I can fix this?

jwz104:
![afbeelding](https://user-images.githubusercontent.com/6162114/205053397-4676bfd1-2688-445b-a07f-a57ccafbcd2c.png)

Looks like I found the place where to error comes from, will look into a way to fix it :)

https://github.com/tpm2-software/tpm2-pytss/blob/51823ef355fdf2cf2320467fa29bcd18ddc9a8b1/src/tpm2_pytss/types.py#L80

---

Forgot to mention that I only have this issue while testing.

jwz104:
[...]

PR #481, opened 2022-12-01
internal: check for sphinx not unittest.mock for docs workaround

whooo:
Might fix https://github.com/tpm2-software/tpm2-pytss/issues/480


src/tpm2_pytss/internal/utils.py
```diff
@@ -18,7 +18,7 @@
 logger = logging.getLogger(__name__)
 
 # Peek into the loaded modules, if mock is loaded, set __MOCK__ to True, else False
-__MOCK__ = "unittest.mock" in sys.modules
+__MOCK__ = "sphinx" in sys.modules
```
williamcroberts, line 21: probably a better hack around it then what I did.

PR #486, opened 2022-12-09
TCTI: Implement bindings to spi-helper

williamcroberts:
Implement bindings to SPI helper TCTI and test.

### DIFF
--- a/docs/tcti.rst
+++ b/docs/tcti.rst
@@ -8,3 +8,7 @@ TCTI
 .. automodule:: tpm2_pytss.TCTILdr
    :members: TCTILdr
    :undoc-members:
+
+.. automodule:: tpm2_pytss.TCTISPIHelper
+   :members: TCTISPIHelper
+   :undoc-members:
--- a/docs/tcti.rst
+++ b/docs/tcti.rst
@@ -2,13 +2,9 @@ TCTI
 ====
 
 .. automodule:: tpm2_pytss.TCTI
-   :members: TCTI
+   :members: TCTI, PyTCTI
    :undoc-members:
 
 .. automodule:: tpm2_pytss.TCTILdr
-   :members: TCTILdr
-   :undoc-members:
-
-.. automodule:: tpm2_pytss.PyTCTI
    :members: TCTILdr
    :undoc-members:
--- a/src/tpm2_pytss/TCTISPIHelper.py
+++ b/src/tpm2_pytss/TCTISPIHelper.py
@@ -1,0 +1,296 @@
+# SPDX-License-Identifier: BSD-2
+
+from .internal.utils import _chkrc, _lib_version_atleast
+from ._libtpm2_pytss import ffi, lib
+from .constants import TSS2_RC, TPM2_RC
+from .TSS2_Exception import TSS2_Exception
+from .TCTI import TCTI
+
+if not _lib_version_atleast("tss2-tcti-spi-helper", "0.0.0"):
+    raise NotImplementedError("Package tss2-tcti-spi-helper not present")
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_sleep_ms(userdata, milliseconds):
+    thiz = TCTISPIHelper._cffi_cast(userdata)
+    if not hasattr(thiz, "on_sleep_ms"):
+        return TSS2_RC.TCTI_RC_NOT_IMPLEMENTED
+    try:
+        thiz.on_sleep_ms(milliseconds)
+    except Exception as e:
+        rc = e.rc if isinstance(e, TSS2_Exception) else TSS2_RC.TCTI_RC_GENERAL_FAILURE
+        thiz._set_last_exception(e)
+        return rc
+
+    return TPM2_RC.SUCCESS
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_start_timeout(userdata, milliseconds):
+    thiz = TCTISPIHelper._cffi_cast(userdata)
+    if not hasattr(thiz, "on_start_timeout"):
+        return TSS2_RC.TCTI_RC_NOT_IMPLEMENTED
+    try:
+        thiz.on_start_timeout(milliseconds)
+    except Exception as e:
+        rc = e.rc if isinstance(e, TSS2_Exception) else TSS2_RC.TCTI_RC_GENERAL_FAILURE
+        thiz._set_last_exception(e)
+        return rc
+
+    return TPM2_RC.SUCCESS
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_timeout_expired(userdata, is_time_expired) -> bool:
+
+    thiz = TCTISPIHelper._cffi_cast(userdata)
+    if not hasattr(thiz, "on_timeout_expired"):
+        return TSS2_RC.TCTI_RC_NOT_IMPLEMENTED
+    try:
+        result = thiz.on_timeout_expired()
+        is_time_expired[0] = result
+    except Exception as e:
+        rc = e.rc if isinstance(e, TSS2_Exception) else TSS2_RC.TCTI_RC_GENERAL_FAILURE
+        thiz._set_last_exception(e)
+        return rc
+
+    return TPM2_RC.SUCCESS
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_spi_acquire(userdata):
+    thiz = TCTISPIHelper._cffi_cast(userdata)
+    if not hasattr(thiz, "on_start_timeout"):
+        return TSS2_RC.TCTI_RC_NOT_IMPLEMENTED
+    try:
+        thiz.on_spi_acquire()
+    except Exception as e:
+        rc = e.rc if isinstance(e, TSS2_Exception) else TSS2_RC.TCTI_RC_GENERAL_FAILURE
+        thiz._set_last_exception(e)
+        return rc
+
+    return TPM2_RC.SUCCESS
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_spi_release(userdata):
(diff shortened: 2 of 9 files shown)


src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,261 @@
...
+
+        self._tcti_mem = ffi.new(f"uint8_t [{size[0]}]")
+        self._opaque_tcti_ctx = ffi.cast("TSS2_TCTI_CONTEXT *", self._tcti_mem)
+
+        _chkrc(
+            lib.Tss2_Tcti_Spi_Helper_Init(self._opaque_tcti_ctx, size, self._callbacks,)
+        )
+
+        self._last_exception = None
+
```
whooo, line 176: `self._last_exception` doesn't seem to be used

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,261 @@
...
+if not _lib_version_atleast("tss2-tcti-spi-helper", "0.0.0"):
+    raise NotImplementedError("Package tss2-tcti-spi-helper not present")
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_sleep_ms(userdata, milliseconds):
+    thiz = TCTISPIHelper._cffi_cast(userdata)
+    if not hasattr(thiz, "on_sleep_ms"):
+        thiz.on_sleep_ms(milliseconds)
+
```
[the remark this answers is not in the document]
whooo: Might be good to catch any exception, otherwise cffi might just dump the traceback to stderr

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,261 @@
...
+    if hasattr(thiz, "on_start_timeout"):
+        thiz.on_start_timeout(milliseconds)
+
+
+@ffi.def_extern()
+def _tcti_spi_helper_timeout_expired(userdata) -> bool:
+    thiz = TCTISPIHelper._cffi_cast(userdata)
+    if hasattr(thiz, "on_start_timeout"):
+        return thiz.on_timeout_expired()
+
```
whooo, line 32: catch exceptions here as well

whooo (changes requested):
tss2-tcti-spi-helper should  be added to FakeVersions in docs/conf.py so it will show up on readthedocs

src/tpm2_pytss/TCTISPIHelper.py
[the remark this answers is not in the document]
williamcroberts: > Might be good to catch any exception, otherwise cffi might just dump the traceback to stderr

That's actually what I wanted. Theirs no mechanism to return errors through the C API, which makes me want to change that before I cut 4.0 of tpm2-tss. The SPI TCTI Helper just retries 100 times and fails with IO error. It would be better to have the ability to error in all the spi callbacks and then we can set last exception and re-throw it in the Python code.

williamcroberts:
I think I need to modify the C library before it ships....

src/tpm2_pytss/TCTISPIHelper.py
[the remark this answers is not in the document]
whooo: One option is to simply catch and pass, no feedback in case of errors, but no random traceback dumps either

williamcroberts:
This PR will now depend on https://github.com/tpm2-software/tpm2-tss/pull/2517

test/test_tcti_spi_helper.py
```diff
@@ -0,0 +1,171 @@
...
+
+import unittest
+from tpm2_pytss import *
+
+from tpm2_pytss.TCTISPIHelper import TCTISPIHelper
+
+# try:
+#     from tpm2_pytss.TCTISPIHelper import TCTISPIHelper
+# except NotImplementedError:
+#     raise unittest.SkipTest("TCTISPIHelper not supported, skipping.")
```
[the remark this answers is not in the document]
williamcroberts: Fixed

src/tpm2_pytss/TCTISPIHelper.py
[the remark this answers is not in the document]
williamcroberts: @whooo 
> One option is to simply catch and pass, no feedback in case of errors, but no random traceback dumps either

I don't get the benefit of that, could you explain? It's moot now, as I updated the API to use return codes everywhere but finalize, however I am curious.

Now the exception framework of the base TCTI class can be used through the set_last_exception and friends methods and then we can also check it in the constructor for the spi helper TCTI and catch errors as well. This allows for exceptions to be handled in the constructor as well as in the TCTI interface calls, like transmit.

docs/tcti.rst
```diff
@@ -10,5 +10,9 @@ TCTI
    :undoc-members:
 
 .. automodule:: tpm2_pytss.PyTCTI
-   :members: TCTILdr
+   :members: PyTCTI
+   :undoc-members:
```
williamcroberts, line 14: I thought commit "docs: fix PyTCTI members" would have fixed the documentation not showing up correctly for PyTCTI but it doesn't seem too, hrmm.

williamcroberts: Ahh that makes sense, auto-module is on the package not the classes. Thanks.

williamcroberts:
@whooo could you do another review pass on this? Your suggestion on docs was right on, thanks.

whooo:
Heh, just as I was going through the code

williamcroberts:
> Heh, just as I was going through the code

Let me know if you spot anything, someone wanted a release and I wanted to pull this into 2.1.0. I just cut 2.1.0-rc0, but theirs no reason we cant follow up with an rc1

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+    @staticmethod
+    def _cffi_cast(userdata):
+        return ffi.from_handle(userdata)
+
+    def on_sleep_ms(self, milliseconds: int) -> None:
+        """Sleeps for a specified amount of time in millisecons.
+
+        This callback is REQUIRED.
+        No errors may occur across this boundary.
+
```
whooo, line 232: Is this still true?

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+            milliseconds(int): The time to sleep.
+        """
+        pass
+
+    def on_start_timeout(self, milliseconds: int) -> None:
+        """Called when a timeout is occurring with the sleep duration in millisecons.
+
+        This callback is REQUIRED.
+        No errors may occur across this boundary.
+
```
whooo, line 243: Same here

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+            milliseconds(int): The time to sleep.
+        """
+        pass
+
+    def on_timeout_expired(self) -> bool:
+        """Called to determine if a timeout is expired.
+
+        This callback is REQUIRED.
+        No errors may occur across this boundary.
+        """
```
whooo, line 254: Same here

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+        pass
+
+    def on_spi_transfer(self, data_in: bytes) -> bytes:
+        """Called to transfer data across the SPI bus.
+
+        This callback is REQUIRED.
+        No errors may occur across this boundary.
+
```
whooo, line 262: Same here

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+            data_in(bytes): The data to send.
+
+        Returns(bytes):
+            The bytes to send.
+
+        Raises:
+            TSS2_Exception: if an error occurs and the RC is returned to the
+                native TCTI.
+        """
+
```
whooo, line 273: Missing pass

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+    def on_finalize(self) -> None:
+        """Called when the TCTI is finalized.
+
+        This callback is OPTIONAL.
+        No errors may occur across this boundary.
+        """
+        pass
```
whooo, line 280: Same here

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def on_spi_acquire(self) -> None:
+        """Called when the SPI bus needs to be acquired for wait states.
+
+        This callback is REQUIRED for WAIT STATES.
+        No errors may occur across this boundary.
+        """
+        pass
```
whooo, line 288: Same here

src/tpm2_pytss/TCTISPIHelper.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def on_spi_release(self) -> None:
+        """Called when the SPI bus needs to be released for wait states.
+
+        This callback is REQUIRED for WAIT STATES.
+        No errors may occur across this boundary.
+        """
+        pass
```
whooo, line 296: Same here

src/tpm2_pytss/TCTISPIHelper.py
williamcroberts: I guess pass isn't needed since they have a docstring, but I guess Ill add it to keep it consistent or should I just delete them?

PR #512, opened 2023-05-18
ci: use tpm2-tools 5.5 by default and tpm2-tss 4.0.1 for coverage job

whooo:


whooo:
@JuergenReppSIT, when you have time, could you have a look at this as well?

.github/workflows/tests.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         tss-version: ['master', '2.4.0', '3.0.0', '3.0.3', '3.2.0', '4.0.0']
         with-fapi: [true]
-        tools-version: ['master']
+        tools-version: ['5.5']
```
JuergenReppSIT, line 13: @whooo Is there a reason to test the tool version 5.5 instead of the master?

whooo: Mostly to allow caching via https://github.com/tpm2-software/tpm2-pytss/pull/387
In the future I'll look into having one test that is tpm2-tss and tpm2-tools master combined

JuergenReppSIT (approved):

PR #525, opened 2023-07-27
tctildr: use Tss2_TctiLdr_GetInfo to lookup tcti backend

baloo:
distros may place TCTI backend to different locations than the system libraries.

This commits makes the tests lookup the backend via `Tss2_TctiLdr_GetInfo` instead of of LD_LIBRARY_PATH.

### DIFF
--- a/src/tpm2_pytss/TCTILdr.py
+++ b/src/tpm2_pytss/TCTILdr.py
@@ -67,3 +67,24 @@ class TCTILdr(TCTI):
 
     def __str__(self):
         return self.name_conf
+
+    @staticmethod
+    def is_available(name=None) -> bool:
+        """Lookup the TCTI and return its availability
+
+        Returns:
+           True if the interface is available
+        """
+        ctx_pp = ffi.new("TSS2_TCTI_INFO **")
+
+        if name is None:
+            name = ffi.NULL
+        elif isinstance(name, str):
+            name = name.encode()
+
+        ret = lib.Tss2_TctiLdr_GetInfo(name, ctx_pp)
+        if ret != lib.TPM2_RC_SUCCESS:
+            return False
+
+        lib.Tss2_TctiLdr_FreeInfo(ctx_pp)
+        return True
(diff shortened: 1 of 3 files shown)


src/tpm2_pytss/TCTILdr.py
```diff
@@ -67,3 +67,19 @@ def name_conf(self):
 
     def __str__(self):
         return self.name_conf
+
+    @classmethod
+    def is_available(cls, name=None):
```
whooo, line 72: You can use `@stadicmethod` and drop `cls` as the method doesn't need it.
Also, please add type annotations and documentation for the method, even if the rest of class doesn't have it.

test/TSS2_BaseTest.py
```diff
@@ -168,7 +168,8 @@ def getSimulator():
                 print(f'Could not find executable: "{sim.exe}"', file=sys.stderr)
                 continue
             try:
-                cdll.LoadLibrary(sim.libname)
+                if not TCTILdr.is_available(sim.libname):
+                    continue
             except OSError as e:
                 print(
```
whooo, line 174: No need for the try...except part as the method doesn't raise any `OSError`

whooo (changes requested):
Please add a unit tests for the method as well

test/test_tcti.py
```diff
@@ -223,6 +223,10 @@ def test_custom_pytcti_finalize_error(self):
         with self.assertRaises(RuntimeError, msg="Bills Error 2"):
             t.finalize()
 
+    def test_is_available(self):
+        self.assertTrue(TCTILdr.is_available("swtpm"))
```
whooo, line 227: You could try using `self.tpm.libname` here, as some tests against older versions of tpm2-tss doesn't use swpm

baloo: I'm trying with `None` as it should gets the default, at least I hope.

baloo: looks like it worked.

whooo (approved):

Issue #527, opened 2023-08-07
Tests test_exec_sign_callback and test_exec_polauth_callback in test_policy.py use prime192v1 key not supported in Fedora


Jakuje:
I am updating the package in Fedora and we have tree tests failing because OpenSSL in Fedora (and and RHEL based system) is for legal reasons missing some curves, namely mostly brainpool, but also the prime192v1 but also sect163r2. This makes the following tests failing:

 * `test_crypto.py`: `test_ecc_bad_curves`: Fails with message `Unsupported key format` instead of `unsupported curve: sect163r2` (I have this patched since [last version](https://src.fedoraproject.org/rpms/python-tpm2-pytss/blob/rawhide/f/python-tpm2-pytss-1.2.0-openssl.patch).
 * `test_policy.py`: `test_exec_polauth_callback` and `test_exec_sign_callback` (using the same key): Our OpenSSL fails to import the prime192v1 private key. The workaround could be to use different curve/key to make it more. So far I will skip this test downstream too, but it would be good to have an upstream and compatible solution that will work everywhere

It also fails to build on 32b arch, but I do not think this is a problem for anyone in 2023.

On s390x it fails waiting for the TPM2 in this test:
```
________________________ TestTCTI.test_spi_helper_good _________________________
[gw0] linux -- Python 3.12.0 /usr/bin/python3
self = <test.test_tcti_spi_helper.TestTCTI testMethod=test_spi_helper_good>
    def test_spi_helper_good(self):
>       MySPITcti()
test/test_tcti_spi_helper.py:142: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/test_tcti_spi_helper.py:37: in __init__
    super().__init__(*args, **kwargs)
../../BUILDROOT/python-tpm2-pytss-2.1.0-1.fc39.s390x/usr/lib64/python3.12/site-packages/tpm2_pytss/TCTISPIHelper.py:210: in __init__
    raise e
../../BUILDROOT/python-tpm2-pytss-2.1.0-1.fc39.s390x/usr/lib64/python3.12/site-packages/tpm2_pytss/TCTISPIHelper.py:202: in __init__
    _chkrc(
[...]

PR #530, opened 2023-08-11
tests: Accept more different errors for sect173r2 curve

Jakuje:
Fixes (partially) #527.


whooo:
This will change testing from https://github.com/tpm2-software/tpm2-pytss/blob/8a7419c730536f1f52ad410820c211e031bbfd77/src/tpm2_pytss/internal/crypto.py#L158 to https://github.com/tpm2-software/tpm2-pytss/blob/8a7419c730536f1f52ad410820c211e031bbfd77/src/tpm2_pytss/internal/crypto.py#L140
So you need to find another curve that the cryptography module supports but not that table in https://github.com/tpm2-software/tpm2-pytss/blob/8a7419c730536f1f52ad410820c211e031bbfd77/src/tpm2_pytss/internal/crypto.py#L30

`openssl ecparam -list_curves and https://cryptography.io/en/2.8/hazmat/primitives/asymmetric/ec/#elliptic-curves could be a good starting point

Jakuje:
> This will change testing from ...

Thats what I am getting on Fedora. Fedora does not know this curve as it is either stripped or hard-disabled (would have to check). It is very discouraged to be used in any practical systems on the link you mentioned. Other option might be changing the expected value based on the OS/Distro/family if it would be more acceptable.

whooo:
> > This will change testing from ...
> 
> Thats what I am getting on Fedora. Fedora does not know this curve as it is either stripped or hard-disabled (would have to check). It is very discouraged to be used in any practical systems on the link you mentioned. Other option might be changing the expected value based on the OS/Distro/family if it would be more acceptable.

Why not just change the certificate/public key to use another curve that will work on Fedora?
The reason is I linked to that specific version of cryptography is that it's the bare minimum I would like to have support for, as it's the used on the the previous Ubuntu LTS release, but that's not a hard requirement.
If you find another curve I can test it on debian unstable at least.

Jakuje:
Right. That might work, but I am not sure how to get the public key in unsupported key format.
My Fedora 37 has the following cuvres (might differ a bit in rawhide, but probably not much):
```
$ openssl ecparam -list_curves
  secp224r1 : NIST/SECG curve over a 224 bit prime field
  secp256k1 : SECG curve over a 256 bit prime field
  secp384r1 : NIST/SECG curve over a 384 bit prime field
  secp521r1 : NIST/SECG curve over a 521 bit prime field
  prime256v1: X9.62/SECG curve over a 256 bit prime field
``` 
I think all of these should work in any debian versions too.

Jakuje:
Just tried with Fedora 38 container, which has slightly more (we stopped stripping brainpool curves), but not sure if it will help:
```
[root@f04b90f8d3e1 /]# openssl ecparam -list_curves
  secp224r1 : NIST/SECG curve over a 224 bit prime field
  secp256k1 : SECG curve over a 256 bit prime field
  secp384r1 : NIST/SECG curve over a 384 bit prime field
  secp521r1 : NIST/SECG curve over a 521 bit prime field
  prime256v1: X9.62/SECG curve over a 256 bit prime field
  brainpoolP256r1: RFC 5639 curve over a 256 bit prime field
  brainpoolP256t1: RFC 5639 curve over a 256 bit prime field
  brainpoolP320r1: RFC 5639 curve over a 320 bit prime field
  brainpoolP320t1: RFC 5639 curve over a 320 bit prime field
  brainpoolP384r1: RFC 5639 curve over a 384 bit prime field
  brainpoolP384t1: RFC 5639 curve over a 384 bit prime field
  brainpoolP512r1: RFC 5639 curve over a 512 bit prime field
  brainpoolP512t1: RFC 5639 curve over a 512 bit prime field
```
>  It would mean that the tests wont run on Fedora, is that acceptable?

I think that should be acceptable. Given that it runs everywhere else and these are just couple of tests to skip, its better that not running them at all.

Jakuje:
Will be resolved with #534 instead.

PR #540, opened 2023-10-06
test: test against python 3.12

whooo:
Python 3.12 is released now, so test against it.


JuergenReppSIT (approved):
@whooo sorry I have overlooked the mail

Issue #548, opened 2023-11-07
docs build failing


williamcroberts:
@whooo the docs build is failing with:
```
        File "/tmp/pip-build-env-kmrvxs08/overlay/lib/python3.12/site-packages/setuptools/_distutils/command/sdist.py", line 336, in _add_defaults_ext
          self.filelist.extend(build_ext.get_source_files())
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "<string>", line 204, in get_source_files
        File "/tmp/pip-build-env-kmrvxs08/overlay/lib/python3.12/site-packages/setuptools/_distutils/cmd.py", line 107, in __getattr__
          raise AttributeError(attr)
      AttributeError: cython_sources
      [end of output]
```

AFAICT, it's an issue with cython per this PR:
  - https://github.com/cython/cython/pull/4498

Which I found through this bug report:
  - https://github.com/readthedocs/readthedocs.org/issues/5297
 
You're a little bit more Python saavy than I, do you see a solution in that? The one that stuck out to me was setting our python version to like 3.6 in the file `.readthedocs.yml` or mocking it.

williamcroberts:
I see you tagged an RC but right now their won't be docs building for the release.

whooo:
Yeah, noticed that as well, found https://github.com/yaml/pyyaml/issues/601
Finally took the time to cleanup docs/requirements.txt a bit

PR #549, opened 2023-11-07
docs: use same requirements as in setup.cfg

whooo:
Cleanup docs/requirements to make it more future proof.
Fixes https://github.com/tpm2-software/tpm2-pytss/issues/548


williamcroberts (approved):
LGTM

Issue #557, opened 2023-12-23
Deprecate tools_encdec


whooo:
The tools_encdec class is hard to maintain and probably not used, so it should be marked as deprecated and removed in a future release.

PR #558, opened 2023-12-23
encoding: add deprecation warning to tools_encdec

whooo:
See https://github.com/tpm2-software/tpm2-pytss/issues/557

### DIFF
--- a/src/tpm2_pytss/encoding.py
+++ b/src/tpm2_pytss/encoding.py
@@ -80,6 +80,7 @@ from .types import (
 )
 import yaml
 import collections.abc
+import warnings
 
 
 class base_encdec(object):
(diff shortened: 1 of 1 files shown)


williamcroberts (approved):
The goal is to get the tools all the way to a standardized YAML MU library within tpm2-tss that can have bindings here as well as used by the tools. I started an auto generator, but never got the time to finish it.

whooo:
> The goal is to get the tools all the way to a standardized YAML MU library within tpm2-tss that can have bindings here as well as used by the tools. I started an auto generator, but never got the time to finish it.

Any plans to support the JSON as well? To work with policies together with tss2-policy

williamcroberts:
> > The goal is to get the tools all the way to a standardized YAML MU library within tpm2-tss that can have bindings here as well as used by the tools. I started an auto generator, but never got the time to finish it.
> 
> Any plans to support the JSON as well? To work with policies together with tss2-policy

Yes, I actually have a draft PR for starting a tpm2-tool for libtss2-policy.

Issue #362, opened 2022-06-20
tox complains about missing library stubs or py.typed.marker


Dvergatal:
Like in subject during testing our code which uses pytss I'm getting this errors:
```bash
device/tpm.py:5: error: Skipping analyzing "tpm2_pytss.ESAPI": module is installed, but missing library stubs or py.typed marker
device/tpm.py:5: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
device/tpm.py:6: error: Skipping analyzing "tpm2_pytss.utils": module is installed, but missing library stubs or py.typed marker
device/tpm.py:7: error: Skipping analyzing "tpm2_pytss.types": module is installed, but missing library stubs or py.typed marker
device/tpm.py:8: error: Skipping analyzing "tpm2_pytss.constants": module is installed, but missing library stubs or py.typed marker
Found 4 errors in 1 file (checked 49 source files)
ERROR: InvocationError for command /work/.tox/mypy/bin/mypy . --strict (exited with code 1)
```
I suppose, that as the link suggests we are `py.typed` file in tpm2-pytss package.

williamcroberts:
What is exactly is this? What's tox (looks like a testing framework)? How do I reproduce? Did you build the package or just grab the code, if you don't pip install it you won't generate the bindings.

Dvergatal:
Our code uses imports from tpm2_pytss. The tpm2_pytss is installed using this command:
```bash
python3 -m pip install tpm2-ptss
```
with newest version 1.2.0. As you can see in my previous log, there is a hint link [missing imports](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports), where you can read about this error. mypy uses tox package for static analysis of our code and due to this test I'm getting these errors.

williamcroberts:
So, per the docs:
```
Mypy will not try inferring the types of any 3rd party libraries you have installed unless they either have declared themselves to be [PEP 561 compliant stub package](https://mypy.readthedocs.io/en/stable/installed_packages.html#installed-packages) (e.g. with a py.typed file) or have registered themselves on [typeshed](https://github.com/python/typeshed), the repository of types for the standard library and some 3rd party libraries.
```

So we need a py.typed file I think... I need information on how to reproduce.

PR #560, opened 2024-01-17
Mypy per module

whooo:


williamcroberts:
@whooo what's the overarching point of these patches? It was lost on me.

whooo:
> @whooo what's the overarching point of these patches? It was lost on me.

To fix https://github.com/tpm2-software/tpm2-pytss/issues/362 and on the way clean up some bugs/bad type annotations

PR #564, opened 2024-02-10
Hopefully next

whooo:
Branch to keep all fixes and new features in one place to ensure that they work together and for easier installation.


williamcroberts:
As you pull features, the CI loop should be testing it, so their should be no need for something like this. You just don't merge something that causes a regression. Can we close this?

williamcroberts:
Whoops, hit close, lets discuss what to do with this.

whooo:
> As you pull features, the CI loop should be testing it, so their should be no need for something like this. You just don't merge something that causes a regression. Can we close this?

I just forgot to mark it as a draft, the CI loop was broken due to changes/bugs in upstream projects

PR #569, opened 2024-03-26
cryptography: fixes for newer version of cryptography

whooo:


whooo:
Merged via https://github.com/tpm2-software/tpm2-pytss/pull/576

Issue #570, opened 2024-04-12
can not install on ubuntu 22.04


josh20170311:
```
ubuntu@ubuntu-vm:~$ uname -a
Linux ubuntu-vm 6.5.0-27-generic #28~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 15 10:51:06 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
ubuntu@ubuntu-vm:~$ pip3 install tpm2-pytss
Defaulting to user installation because normal site-packages is not writeable
Collecting tpm2-pytss
  Using cached tpm2-pytss-2.2.1.tar.gz (208 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
 
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [61 lines of output]
      /tmp/pip-build-env-od6zhajc/overlay/local/lib/python3.10/dist-packages/setuptools_scm/_integration/setuptools.py:30: RuntimeWarning:
      ERROR: setuptools==59.6.0 is used in combination with setuptools_scm>=8.x
 
      Your build configuration is incomplete and previously worked by accident!
      setuptools_scm requires setuptools>=61
 
      Suggested workaround if applicable:
       - migrating from the deprecated setup_requires mechanism to pep517/518
         and using a pyproject.toml to declare build dependencies
         which are reliably pre-installed before running the build tools
 
        warnings.warn(
      adding path: /tmp/pip-install-8pyig6_n/tpm2-pytss_935acbc0407c4f19803363a2c5866332/scripts
      Traceback (most recent call last):
        File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
          main()
        File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py", line 164, in prepare_metadata_for_build_wheel
          return hook(metadata_directory, config_settings)
        File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 174, in prepare_metadata_for_build_wheel
          self.run_setup()
[...]

PR #571, opened 2024-04-20
Fixes for tpm2 tss master

whooo:
One revert for as the issue has been fixed in tpm2-tss and a fix for changes to some defines.

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/570


whooo:
Merged via https://github.com/tpm2-software/tpm2-pytss/pull/576

PR #534, opened 2023-08-16
test: skip tests if ECC curves aren't supported.

whooo:
Some Linux distributions have a limited set of supported ECC curves in openssl, so instead of finding a curve that fits everyone skip certain tests on those systems.

Fixes https://github.com/tpm2-software/tpm2-pytss/issues/527

PR #575, opened 2024-05-28
Compatibility with python-cryptography 42

Jakuje:
based on initial changes proposed in #534 which did not work for the old cryptography, but this should work for both.


Jakuje:
for other tests to work, i think some more changes such as #571 are needed.

whooo:
@Jakuje, could you rebase on master and do a push?

Jakuje:
> @Jakuje, could you rebase on master and do a push?

done.

whooo (approved):

PR #577, opened 2024-05-28
really drop python<=3.7 support

kloczek:
Filter all code over `pyupgrade --py38-plus`.


williamcroberts (commented):
LGTM, just need to re-run it on the latest master branch.

PR #585, opened 2024-07-03
Skip test using P-192 when not available

Jakuje:
and fix the typo in the curve name in other places this is used.

### DIFF
--- a/test/test_cryptography.py
+++ b/test/test_cryptography.py
@@ -297,14 +297,14 @@ class TestCryptography(TSS2_EsapiTest):
         )
         privkey = tpm_ecc_private_key(self.ectx, handle)
 
-        peer_key = ec.generate_private_key(ec.SECP192R1())
+        peer_key = ec.generate_private_key(ec.SECP384R1())
         peer_public_key = peer_key.public_key()
 
         with self.assertRaises(ValueError) as e:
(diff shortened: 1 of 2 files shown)


whooo:
Just so I know, is this related to missing ECC curves in some distributions (due to patents)?

Jakuje:
I do not know all the background, but the reason for excluding the P-192 I remember was that its too small to be secure.

Fedora spec file is not much more verbose either:

https://src.fedoraproject.org/rpms/openssl/blob/rawhide/f/openssl.spec#_65

whooo:
You should be able to switch to another curve available in the openssl/cryptography, the check is done on cryptography types, so the curve doesn't need to be available in the TPM simulator

Jakuje:
Indeed! You are right. This fails early before getting to the tpm2 so we could use any other that should be widely available, such as `ec.SECP384R1`. Let see if it will work in all the cases. Verified it works on Fedora rawhide (in mock).

whooo:
The build error is fixed on master now, could you rebase and push?

Jakuje:
Reran, but there is still some likely intermittent failure.

whooo (approved):

whooo:
Thanks for your continuing contributions!

PR #586, opened 2024-07-13
scripts: fix regex for header preparation

whooo:
There are more upstream header file changes, so make sure we handle them correctly.


whooo:
Merged in https://github.com/tpm2-software/tpm2-pytss/pull/587

PR #594, opened 2024-08-29
Fix typo regarding format string in FAPI.py

Semnodime:


whooo:
@Semnodime, please do git commit -s --amend to agree to DCO

PR #599, opened 2024-09-13
ESAPI: add context manager to flush ESAPI handles.

### LINKED ISSUE #595: flush parent handle durint create and load
the parent context isn't flushed during a create and load commands  cleanly which causes issues in using pytss with a swtpm (no resource mgr, etc).  

suggestion is to add in the flush commands shown below to `tpm2_pytss/tsskey.py`.

i can submit a small pr if needed

```python
    def load(self, ectx, password=None):
        if not password and not self.empty_auth:
            raise RuntimeError("no password specified but it is required")
        elif password and self.empty_auth:
            warnings.warn("password specified but empty_auth is true")
        phandle = self._getparent(ectx, self.public.publicArea.type, self.parent)
        handle = ectx.load(phandle, self.private, self.public)
        ectx.tr_set_auth(handle, password)
        ectx.flush_context(phandle)  # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        return handle

    @classmethod
    def create(cls, ectx, template, parent=lib.TPM2_RH_OWNER, password=None):
        insens = TPM2B_SENSITIVE_CREATE()
        emptyauth = True
        if password:
            insens.sensitive.userAuth = password
            emptyauth = False
        phandle = cls._getparent(ectx, template.type, parent)
        private, public, _, _, _ = ectx.create(
            parent_handle=phandle,
            in_sensitive=insens,
            in_public=TPM2B_PUBLIC(publicArea=template),
            outside_info=TPM2B_DATA(),
            creation_pcr=TPML_PCR_SELECTION(),
        )
        ectx.flush_context(phandle)        # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        return cls(private, public, emptyauth, parent)
```

whooo:
With this is possible to explicity flush transient handles. Useful with temporary handles, for example a primary key handle when just loading a key and when there are no resource manager available.

Fixes #595

### DIFF
--- a/src/tpm2_pytss/ESAPI.py
+++ b/src/tpm2_pytss/ESAPI.py
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: BSD-2
+
+import contextlib
 from .types import *
 from .constants import *
 from .internal.utils import (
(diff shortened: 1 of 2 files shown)


whooo:

PR #600, opened 2024-09-14
encoding: remove tools support as promised

### LINKED ISSUE #557: Deprecate tools_encdec

whooo:
Closes #557


whooo:

PR #601, opened 2024-09-20
setup: bump required minimal version of cryptography to 43.0.0

whooo:
Also remove some code that is uneeded after the bump.


whooo:

PR #602, opened 2024-09-23
add fixes for some CodeQL warnings

whooo:
* Drop unused import
* Use __le__ instead of __lte__ for TSS2Version
* Drop _port from BaseTPMSimulator subclasses


whooo:

PR #605, opened 2024-10-02
ci: add ruff linter

whooo:
Only checks for unused imports for now, more rules can be added later.


whooo:

PR #607, opened 2024-10-14
ci: fixes for latest ubuntu-latest image

whooo:


whooo:

PR #608, opened 2024-10-14
setup/ci: add python 3.13 and drop python 3.8

whooo:
3.8 is no longer supported upstream and 3.13 has been released.


whooo:

PR #615, opened 2024-12-15
setup.py: Explicilty use C99 for preprocessing

salahcoronya:
GCC 15 uses -std=c23 by default, which causes pycparser to error out with "pycparser.plyparser.ParseError:
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/stddef.h:450:31: before: nullptr_t". As pcyparser only supports completely C99, this shouldn't be an issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -189,6 +189,7 @@ class type_generator(build_ext):
             pdata = preprocess_file(
                 header_path,
                 cpp_args=[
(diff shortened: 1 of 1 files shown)


whooo (approved):

whooo:
Thanks for your contribution!

PR #618, opened 2025-02-03
scripts: fix regexp for poll stuff

whooo:
Support for zephyr in tpm2-tss broke the removal of poll conditionals. So make the matching more flexible.


whooo:
Merged via https://github.com/tpm2-software/tpm2-pytss/pull/619

PR #620, opened 2025-02-09
scripts: use raw strings for regular expressions

whooo:
Use raw strings for all regular expressions.
This fixes the related syntax warnings during the build process.


whooo:
Merged in https://github.com/tpm2-software/tpm2-pytss/pull/627

Issue #625, opened 2025-02-15
Add missing 1.59 commands, structures and constants


whooo:
Some are missing, so make sure to add them

whooo:
Fixed in https://github.com/tpm2-software/tpm2-pytss/pull/630

PR #628, opened 2025-02-23
Missing constants and functions

whooo:
Adds the act_set_timeout and certify_x509 commands and some missing constants

Closes https://github.com/tpm2-software/tpm2-pytss/issues/625


whooo:
Merged via https://github.com/tpm2-software/tpm2-pytss/pull/630

PR #631, opened 2025-03-06
setup.py: Preprocess using CC environmental variable

salahcoronya:
This assists in cross-compiling and choice in choosing which CPP implementation is used (gcc vs. clang).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,14 @@ from textwrap import dedent
 site.ENABLE_USER_SITE = "--user" in sys.argv[1:]
 
 
+def cpp_path():
+    return os.environ.get("CC", "cc")
+
+
+def cpp_args(args=[]):
+    return ["-E"] + args
+
+
 class type_generator(build_ext):
     cares = set(
         (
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -21,6 +21,19 @@
...
+
+
+def cpp_args(args=[]):
+    cc = os.environ.get("CC")
+
+    if cc is not None:
+        return ["-E"] + args
+    else:
+        return args
+
```
whooo, line 35: Can't we just pass "-E" directly to cpp_args regardless of if $CC is set?

salahcoronya: clang's cpp implementation will complain:

cpp: warning: argument unused during compilation: '-E' [-Wunused-command-line-argument]

whooo: I can live with the warning, but I'll leave it to you

salahcoronya: The less unnecessary warnings the better, but its not a big deal. If you dislike the optional parameter I can remove it - but then you you want to push the option directly into the preprocess_file?

setup.py
```diff
@@ -21,6 +21,19 @@
 site.ENABLE_USER_SITE = "--user" in sys.argv[1:]
 
 
+def cpp_path():
+    return os.environ.get("CC", "cpp")
```
mgorny, line 25:
```suggestion
    return os.environ.get("CC", "cc")
```

Then you don't need special conditions. Though normally it's `cc -e` as a single argument, but I guess pycparser has a really bad design.

whooo (dismissed):

setup.py
whooo: Lets go with your current version

setup.py
mgorny: Oh, and remember that `CC` can actually contain options, so unless pycparser can handle that, it would probably make sense to `shlex.split()` it first.

whooo: > Then you don't need special conditions. Though normally it's `cc -e` as a single argument, but I guess pycparser has a really bad design.

@salahcoronya, how do you feel about this? I'm happy with the patch as is now, but do you want to change it?

salahcoronya: I just pushed the improved version moments before you commented.

setup.py
```diff
@@ -21,6 +22,14 @@
 site.ENABLE_USER_SITE = "--user" in sys.argv[1:]
 
 
+def cpp_path():
+    return shlex.split(os.environ.get("CC", "cc"))[0]
+
+
+def cpp_args(args=[]):
+    return ["-E"] + shlex.split(os.environ.get("CC", ""))[1:] + args
+
```
whooo, line 31: I rather avoid the complexity if there isn't a practical use case.
So please drop the shlex part, so something like:
```python
if os.environ.get("CC", None):
    args = ["-E",] + args
return args
```

And drop the shlex in `cpp_path`

whooo (approved):

PR #635, opened 2025-04-03
Fix ESAPI.flush_handle

paulbartell:
The logic in the flush_handle function of ESAPI.py is incorrect. 

Rather than comparing handles against the 24-bit left shifted TPM2_HT type, It uses the TPM2_HR type. Thus all calls to flush_handle prior to this change set never actually flushed the given handle.


whooo:
Fix the style check, the other issues is due to changes in upstream, I'll fix them

Issue #639, opened 2025-05-03
__typeof__ causes build failure


williamcroberts:
On F42 Docker container installing via pip fails:
```bash
pip3 install tpm2-pytss
```

With:
```bash
 File "/tmp/pip-build-env-8noox3r5/overlay/lib/python3.13/site-packages/pycparser/ply/yacc.py", line 331, in parse
          return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
                 ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-8noox3r5/overlay/lib/python3.13/site-packages/pycparser/ply/yacc.py", line 1199, in parseopt_notrack
          tok = call_errorfunc(self.errorfunc, errtoken, self)
        File "/tmp/pip-build-env-8noox3r5/overlay/lib/python3.13/site-packages/pycparser/ply/yacc.py", line 193, in call_errorfunc
          r = errorfunc(token)
        File "/tmp/pip-build-env-8noox3r5/overlay/lib/python3.13/site-packages/pycparser/c_parser.py", line 1945, in p_error
          self._parse_error(
          ~~~~~~~~~~~~~~~~~^
              'before: %s' % p.value,
              ^^^^^^^^^^^^^^^^^^^^^^^
              self._coord(lineno=p.lineno,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                          column=self.clex.find_tok_column(p)))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-8noox3r5/overlay/lib/python3.13/site-packages/pycparser/plyparser.py", line 67, in _parse_error
          raise ParseError("%s: %s" % (coord, msg))
      pycparser.plyparser.ParseError: /usr/lib/gcc/x86_64-redhat-linux/15/include/stddef.h:465:31: before: nullptr_t
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for tpm2-pytss
```

I think it's because of the `__typeof__`

whooo:
It should already be fixed in master, just not in any release yet

williamcroberts:
> It should already be fixed in master, just not in any release yet

[...]

PR #640, opened 2025-05-03
setup: fix parser error on __typeof__

williamcroberts:
Fix building the wheel on newer versions of GCC 15's stdlib that uses __typeof__, which is not understood by pycparser.

Fixes: #639


williamcroberts:
Confirmed fixed on master.

PR #643, opened 2025-05-18
cryptograpy: add copy dunder for private keys

whooo:
cryptography >= 45.0.0 requires the copy dunder for private key implementations.


williamcroberts (approved):
LGTM.

PR #657, opened 2025-06-07
types: add marshal/unmarshal methods for unions.

### LINKED ISSUE #651: Cannot marshal unions directly
One cannot marshal unions directly since they need a type selector. Perhaps make the marshal routine smart enough to know if it's in a union and need an extra argument for the selector value or seperate the union marshal out to it's own thing that union types can inherit from.

```python
b = TPMU_PUBLIC_ID(rsa=b"\x00" * 256).marshal()
  File "<python-input-3>", line 1, in <module>
    b = TPMU_PUBLIC_ID(rsa=b"\x00" * 256).marshal()
  File "/usr/local/lib64/python3.13/site-packages/tpm2_pytss/types.py", line 222, in marshal
    _chkrc(mfunc(_cdata, buf, 4096, offset))
           ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Tss2_MU_TPMU_PUBLIC_ID_Marshal expected 5 arguments, got 4
>>>
```

whooo:
To marshal/unmarshal an union a selector is needed. So add methods which takes the selector as an argument.

Closes #651

### DIFF
--- a/src/tpm2_pytss/types.py
+++ b/src/tpm2_pytss/types.py
@@ -39,6 +39,7 @@ from tpm2_pytss.constants import (
     TPM2_ECC_CURVE,
     TPM2_SE,
     TPM2_HR,
+    TPM_INT_MU,
 )
 from typing import Union, Tuple, Optional
 import sys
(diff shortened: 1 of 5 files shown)


src/tpm2_pytss/types.py
```diff
@@ -541,7 +541,52 @@ def __iter__(self):
         return TPML_Iterator(self)
 
 
-class TPMU_PUBLIC_PARMS(TPM_OBJECT):
+class TPMU_OBJECT(TPM_OBJECT):
```
williamcroberts, line 544: Probably need todo #660 and massage the request and refactor it so we have:
TPM_OBJECT becomes TPM_BASE_OBJECT which is a base class without marshaling support
TPM_OBJECT would extend TPM_BASE_OBJECT and add marshalling support/unmarshalling
TPMU_OBJECT would extend TPM_BASE_OBJECT and add marshalling support/unmarshalling with the extra parameter for the union

williamcroberts (changes requested):

src/tpm2_pytss/types.py
whooo: It doesn't work for TPM2_HANDLE as it isn't a structure, so that we would need to keep the M/U methods for TPM2_HANDLE

williamcroberts: > It doesn't work for TPM2_HANDLE as it isn't a structure, so that we would need to keep the M/U methods for TPM2_HANDLE

Can't you just detect that the underlying type is a scalar and its size and invoke the correct `Tss2_MU_UINT32_Marshal` and unmarshal routine, where 32 is replaced with the scalar's size?

williamcroberts: Can't that be smartened up to say if scalar do this else do existing behavior?

whooo: I think that would just make the code more complicated and unreadable.
Having one separate mixin class for integer based types would remove the complexity, it could be shared between TPM2_HANDLE, the different constants and the (U)INT* stuff in command_parser

whooo: It's fixed now

PR #663, opened 2025-07-23
Do not require wheel for building

hroncok:
- current version of setuptools (70.1+) does not need wheel at all
 - older versions of setuptools would fetch wheel when building wheels (but not sdists)


whooo:
Looks good to me, and the checks passes, but please do sign off the commit to make the DCO-check happy :-)

hroncok:
Also I am not entirely sire what I would sign off. I did not easily locate any DCO in the README etc. (But I am using my phone to check, so perhaps I did not look good enough.)

whooo:
I see that we don't have anything related to DCO in the documentation, I'll add it in the future, but see https://developercertificate.org/ for now.

PR #674, opened 2025-09-21
ESAPI: make some policy commands arguments optional

### LINKED ISSUE #649: policy command enhancements
in `policy_secret`, for example, the nonce and the expiration are required. We can make a nonce of `None` get the current nonce on the digest? `None` for the expiration could mean to max it out?

This is just thoughts off the top of my head to make the API easier, but it may not be right or the best way.

whooo:
Use the max expiration value by default.
Use an empty buffer for nonce_tpm, cp_hash_a and policy_ref by default.

Closes #649


whooo:
@williamcroberts, if this looks good I'll do a RC for 3.0.0 after it's merged

whooo:
> > @williamcroberts, if this looks good I'll do a RC for 3.0.0 after it's merged
> 
> What's the API change were making?

Excluding this, the tools encoder/decoder has been dropped.

williamcroberts:
> > What wasn't backwards compatible, did we have to move an argument I didn't see?
> 
> Yes, most expiration, nonce_tpm, policy_ref and cp_hash_a has been moved to make them optional

Yeah that's what I imagined, the diff output reviewing it on my phone I didn't see that. Thanks for all the hard work on this, I appreciate it

PR #677, opened 2025-10-13
ci: test against python 3.14 and drop 3.9

whooo:
Python 3.14 has been released, so add it to the test matrix.
And drop 3.9.


williamcroberts (approved):
Yeah, I agree, looks like 3.9 is EOL end of October. Thanks.

PR #680, opened 2025-11-12
scripts: update regex for current head of tpm2-tss

whooo:
There have some more changes to the tpm2-tss headers, so we need to update the regular expressions.

### DIFF
--- a/scripts/prepare_headers.py
+++ b/scripts/prepare_headers.py
@@ -10,6 +10,8 @@ import textwrap
 
 
 def remove_common_guards(s):
+    # merge split lines
+    s = s.replace("\\\n", "")
 
     # Remove includes and guards
     s = re.sub(r"#ifndef.*", r"", s)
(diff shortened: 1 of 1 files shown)


williamcroberts (approved):
This looks like it was painful, I remember doing the first pass and it sucked!

williamcroberts:
> > This looks like it was painful, I remember doing the first pass and it sucked!
> 
> Oh yes, reached a point where I was thinking of writing a preprocesser myself

You know, that's not an awful idea. Just pass everything to gcc -E to run the preprocessor and then we don't have to worry about this.

whooo:
> 
> You know, that's not an awful idea. Just pass everything to gcc -E to run the preprocessor and then we don't have to worry about this.

I recall trying that previously, I don't remember why it didn't work out, might have been that all the constants was removed

williamcroberts:
> > You know, that's not an awful idea. Just pass everything to gcc -E to run the preprocessor and then we don't have to worry about this.
> 
> I recall trying that previously, I don't remember why it didn't work out, might have been that all the constants was removed

They would be, darn it. Chatgpt says this works:
gcc -E -dD header.h

There's a lot of -d options we could use to probably get whatever we need.

PR #685, opened 2026-02-22
ci: use libtpms release instead of master.

whooo:
Changes in upstream master sometimes breaks our jobs. So use an release to avoid issues in upstream.


williamcroberts (approved):
Seems reasonable and prudent, LGTM

PR #688, opened 2026-04-26
cryptography: fix support for cryptography 47.

whooo:
The abstract classes now require deepcopy support, so add that.


whooo:
Merged in https://github.com/tpm2-software/tpm2-pytss/pull/689

PR #690, opened 2026-05-02
ci: test with free-threaded python

whooo:


src/tpm2_pytss/cryptography.py
```diff
@@ -263,6 +263,14 @@ def __copy__(self) -> "tpm_rsa_private_key":
             ectx=self._ectx, handle=self._handle, session=self._session
         )
 
+    def __deepcopy__(self, memo: dict) -> "tpm_rsa_private_key":
```
williamcroberts, line 266: Does having the shallow copy limitation have any foreseeable issues?

Maybe we should add documentation about thread safety and mention this along with needing proper serialization around objects that use ESAPI/FAPI.

williamcroberts (approved):
I won't hold this up, just questions.

src/tpm2_pytss/cryptography.py
whooo: > Does having the shallow copy limitation have any foreseeable issues?
> 
Not that I'm aware of, I suspect it isn't really used, just required by the cryptography module.

> Maybe we should add documentation about thread safety and mention this along with needing proper serialization around objects that use ESAPI/FAPI.
This is mostly a start, but my understanding is that unless a C module doesn't report support for free threading, the GIL will be enabled.
But I was thinking of adding more explicit test (for example running the tests in threads).

PR #691, opened 2026-05-10
ci: use swtpm releases instead of master

whooo:
As we already use specific libtpms releases, do the same for swtpm. Recent commits in swtpm requires unreleased versions of libtpms.

Also include small typo fix.


williamcroberts:
Lgtm

williamcroberts (approved):

PR #694, opened 2026-06-30
fix(fapi): keep callback output buffers alive

hyperfinitism:
This pull request fixes FAPI callback output buffer lifetime handling.

The auth and sign callbacks allocate CFFI buffers and pass their pointers back to FAPI. Those buffers need to remain alive after the Python callback returns, because FAPI consumes the returned pointers later in the call path.

This PR stores callback output cdata on the callback metadata object so that the buffers remain referenced for the required lifetime.

### Tests

Tested with Python 3.14.6 free-threaded (`3.14t`):

```bash
python3.14t -m pytest -q \
  test/test_fapi.py::TestFapiECC::test_policy_signed \
  test/test_fapi.py::TestFapiRSA::test_policy_signed \
  test/test_fapi.py::TestFapiRSA::test_policy_action \
  test/test_fapi.py::TestFapiRSA::test_sign \
  test/test_fapi.py::TestFapiRSA::test_get_tpm_blobs \
  test/test_fapi.py::TestFapiRSA::test_encrypt_decrypt
```

Before this fix, these FAPI PolicySigned tests would occasionally fail with a `TPM_RC_SIGNATURE` error:

```console
FAILED test/test_fapi.py::TestFapiRSA::test_policy_signed
tpm:parameter(5):the signature is not valid
```

After this fix, the same test set passes:

```console
6 passed
```

### DIFF
--- a/src/tpm2_pytss/FAPI.py
+++ b/src/tpm2_pytss/FAPI.py
@@ -124,6 +124,7 @@ class _FAPI_CB_UDATA:
         self.cur_exc = None
         self.udata = udata
         self.cb = cb
+        self.cdata_ref = None
 
 
 @ffi.def_extern()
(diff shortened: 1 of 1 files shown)


whooo:
Sorry for the delay, will get to this in the coming days.
Nice find!

src/tpm2_pytss/FAPI.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, cb, udata):
         self.cur_exc = None
         self.udata = udata
         self.cb = cb
+        self.cdata_refs = []
```
whooo: Any specific reason it's a list?

hyperfinitism: No specific current FAPI callback needs multiple callback-owned buffers. The list was intended as generic storage for callback-created cdata references, but today auth and sign each only return one buffer.

I agree a single field would be more precise for the current API surface.

whooo: Would you mind changing that? Otherwise the change looks good

whooo (approved):

whooo:
Thanks for your contribution!

PR #698, opened 2026-08-05
fix(internal): use block size instead of key size for sym encryption

gomesj:
For modes like CFB, OFB and CBC, python cryptography library expects IV size to be equal to the block size.
AES uses a fixed block size of 128 bits (and so does Camellia from a quick search).
This creates a size difference between key and block size when using any key size higher than 128 bits.
While this might not cover expectations for all possible modes, changing the IV size from key length to block size should fix compability with 256 bits key size for the modes currently supported in the internal crypto library (currently mainly CFB due to `_symdef_to_crypt`).


whooo:
Just a small nitpick thing, could you change "fix(internal)" in the commit to "internal/crypto"?
Just to keep it inline with (most) other commit messages.
Otherwise it looks good to me

gomesj:
Sure thing, changed the commit message prefix to "internal/crypto"

whooo (approved):

whooo:
Thanks for your contribution!

